PR #2, opened 2018-10-30
chore(examples): Make examples self-contained

httpete-ire:
When I first looked at the examples I didnt know how I could start playroom using the examples. 

I copied Next.js's approach to examples where each example has its own README and package.json.

I have only converted the `material-ui` and `reakit` for now but I will happily convert the other examples to use the same structure.


markdalgleish:
Thanks for the PR! Sorry for not getting back to you earlier.

This looks great, and would definitely help clean up the project's dev dependencies too.

Just for consistency, would you mind using Yarn for the examples too?

yarn.lock
```diff
@@ -5,14 +5,12 @@
 "@babel/code-frame@^7.0.0":
   version "7.0.0"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.0.0.tgz#06e2ab19bdb535385559aabb5ba59729482800f8"
-  integrity sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==
```
markdalgleish: Which version of Yarn are you using? Looks like it's removing all of these lines.

httpete-ire: I'm on yarn version `1.9.4`

httpete-ire:
No problem @markdalgleish, I'm going to clean up the other examples too

yarn.lock
markdalgleish: Could you maybe try updating to the latest version of Yarn and regenerating the lockfile?

.gitignore
```diff
@@ -1,5 +1,6 @@
 node_modules/
 package-lock.json
+examples/**/yarn.lock
```
markdalgleish: Unless I'm missing something, we'd want to keep the lock file for the examples too, wouldn't we?

yarn.lock
httpete-ire: Upgraded to `yarn 1.12.3` and regenerated the `lockfile`

.gitignore
httpete-ire: 👍

markdalgleish:
Thanks for the PR! Glad we finally got this one through 🎉

PR #4, opened 2018-11-07
feat: Autocomplete via propTypes, auto close tags, quotes and braces

markdalgleish:
This adds initial autocomplete via CodeMirror's `show-hint` and `xml-hint` addons. This works because the `jsx` mode internally switches to `xml` mode when writing JSX tags.

We generate the set of tags, prop names and supported values from the `propTypes` object thanks to [parse-prop-types](https://github.com/diegohaz/parse-prop-types). Right now, when suggesting values, this only supports string values that were defined with `PropTypes.oneOf([...])`, but it's much better than nothing.

In the prop suggestions, we're deliberately omitting `children` (since they shouldn't be provided as attributes) and `className` (since Playroom is designed to work with components, not style sheets).

Also, to help streamline the editing experience, we're now automatically closing JSX tags, quotes and braces, and converting `Tab` key presses into spaces.

screen shot 2018-11-08 at 7 23 25 am

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -55,6 +55,11 @@ const showUsage = () => {
     ? path.resolve(cwd, args.config)
     : await findUp('playroom.config.js', { cwd });
 
+  if (!configPath) {
+    console.error('Please add a playroom.config.js to the root of your project.');
+    process.exit(1);
+  }
+
   const config = require(configPath);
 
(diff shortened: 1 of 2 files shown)


markdalgleish:
Just merged from master and fixed conflicts. Ready for review 👍🏻

PR #7, opened 2018-11-08
feat: Make code editor resizable and detachable

jackhurley23:
This PR implements both a resizable editor, and a detachable editor.

The resizing is implemented using `re-resizable` and the detachable editor is implemented with react portals.  See [here](https://hackernoon.com/using-a-react-16-portal-to-do-something-cool-2a2d627b0202) for external window/portal inspiration.

Demo:
https://cl.ly/909106ad4a23

Limitations:
The new window needs to be focussed by the user before the code renders.
The window.open gets blocked by the browser, not sure if there are approaches here to avoid this.
Basic testing in latest Chrome/Safari/FF


package.json
```diff
@@ -83,8 +84,10 @@
     "webpack-merge": "^4.1.4"
   },
   "devDependencies": {
+    "@babel/plugin-transform-react-constant-elements": "^7.0.0",
```
markdalgleish: Does this dependency still need to be here?

PR #8, opened 2018-11-08
ci(PullApprove): Validate PR titles with PullApprove

markdalgleish:


michaeltaranto:
Thoughts on leveraging the [`extends`](https://www.pullapprove.com/docs/config/extends/) functionality now, given this is a second consumer of the v3 api?

markdalgleish:
We definitely should—that would be awesome. I'd like to get this in as a priority first, though.

PR #14, opened 2018-11-17
fix: Open correct URL if custom port is supplied

pascalduez:
- With a config key `port`

### DIFF
--- a/lib/start.js
+++ b/lib/start.js
@@ -17,12 +17,13 @@ module.exports = (config, callback) => {
 
   const compiler = webpack(webpackConfig);
   const devServer = new WebpackDevServer(compiler, webpackDevServerConfig);
+  const { port = 9000 } = config;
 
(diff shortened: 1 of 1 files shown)


markdalgleish:
Thanks for the PR!

For now, to keep things simple, I'd actually like to avoid extra command line args and aliases, sticking with just a config file—especially since the config is JavaScript, so you can implement your own shortcuts in the meantime (e.g. `port: process.env.PORT || 9000`).

Would you be able to simplify this to a config-only setting? It's actually already there, I just accidentally hard-coded the default value in the `opn` call, so this is potentially a one-line fix which could be a patch release.

markdalgleish:
Fixed, and released as v0.3.1. Thanks for the PR! 😄

pascalduez:
@markdalgleish Thanks for playroom! ;)

PR #15, opened 2018-11-17
feat: Allow configuring default open in browser

pascalduez:
- Add a new `no-open` command line option

Sadly I couldn't go with a more straightforward option like `--open` since `command-line-args` disallow configurable Boolean options like `--open=false`. See https://github.com/75lb/command-line-args/issues/71

The other solution is `--open=on|off` 

Or using `--open` but changing the current behavior and not forcing it by default.   
My preference :)


markdalgleish:
Thanks for this PR too! 😄 

Similarly to https://github.com/seek-oss/playroom/pull/14, would you be able to implement this as an `openBrowser` option in the config file, with a default value of `true`? I'd still like to keep the default behaviour, since it's a much better experience for newcomers to the project who won't know what URL to open.

markdalgleish:
Published as v0.4.0. Thank you! 😄

PR #28, opened 2018-11-29
test(cypress): Add initial Cypress smoke test

markdalgleish:


cypress/support/utils.js
```diff
@@ -0,0 +1,16 @@
+export const typeCode = code =>
```
jackhurley23: Nice 🎉

PR #30, opened 2018-11-30
build(Update webpack plugin): Update html-webpack-plugin to 4

hipstersmoothie:
Storybook depends on html-webpack-plugin@4 and i interface with it through plugins. playroom
depends on html-webpack

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "dedent": "^0.7.0",
     "find-up": "^3.0.0",
     "friendly-errors-webpack-plugin": "^1.7.0",
-    "html-webpack-plugin": "^3.2.0",
+    "html-webpack-plugin": "^4.0.0-beta.4",
     "less": "^3.8.1",
     "less-loader": "^4.1.0",
     "localforage": "^1.7.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4364,9 +4364,10 @@ html-entities@^1.2.0:
   version "1.2.1"
   resolved "https://registry.yarnpkg.com/html-entities/-/html-entities-1.2.1.tgz#0df29351f0721163515dfb9e5543e5f6eed5162f"
 
-html-minifier@^3.2.3:
+html-minifier@^3.5.20:
   version "3.5.21"
   resolved "https://registry.yarnpkg.com/html-minifier/-/html-minifier-3.5.21.tgz#d0040e054730e354db008463593194015212d20c"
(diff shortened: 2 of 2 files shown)


markdalgleish:
Thanks for the PR!

Can you show me an example of how you interface with html-webpack-plugin? Why do you need this in a Playroom environment?

hipstersmoothie:
I dont explicitly interface with it with playroom. I use it in storybook to inject some styles into the manager https://github.com/hipstersmoothie/html-webpack-insert-text-plugin. I could do the same for playroom but i've found easier ways to do that.

My problem here is that since playroom depends on an older version (which gets resolved first) leading to the wrong html-webpack-plugin in my node_modules.

hipstersmoothie:
(Note: I am using yarn resolutions but it doesn't seem to be working in monorepos)

markdalgleish:
This sounds like an issue with your package manager. Unless I'm missing something, we shouldn't have to change our dependencies to make an unrelated tool resolve its own dependencies correctly.

hipstersmoothie:
Would you be okay with the update once it's out of beta? I can't see why updating the dep is bad, even if it's for unrelated reasons.

hipstersmoothie:
I see playroom and storybook as two tools that go together. this dep mismatch is causing problems so why not fix it? Should i be running my playroom in another repo?

hipstersmoothie:
like i can maintain my own fork if you don't want to accept the change. 

I've looked at the code and you aren't using the plugin in any special ways so nothing will break

https://github.com/seek-oss/playroom/blob/406f2aa73523b723bab90364e7193f15ca2dedcf/lib/makeWebpackConfig.js#L95

markdalgleish:
Yeah, once it's out of beta I'd be keen to upgrade.

Are you able to dig into the issue a bit further and provide some more details? I'm not convinced this is an issue with Playroom. Storybook and Playroom should be able to run separate versions of html-webpack-plugin without conflicting with each other, since they don't interact with each other at all.

hipstersmoothie:
They are running seperate version of html-webpack-plugin. My node_modules end up looking like this

node_modules:

html-webpack-plugin@3.2.0
playroom
storybook
   node_modules
      html-webpack-plugin@4
html-webpack-insert-plugin (depends on html-webpack-plugin@4)

as you can see when html-webpack-insert-plugin runs it will import html-webpack-plugin@3.2.0 playroom's html-webpack-plugin instead of storybooks

markdalgleish:
Maybe try installing the v4 beta as a direct dependency of your project?

hipstersmoothie:
I ended up using nohost and it works now

```
    "nohoist": [
      "**/playroom"
    ]
```

thanks for your time!

PR #35, opened 2018-12-06
feat: Add initial TypeScript autocomplete support

markdalgleish:
Before starting webpack, if we detect a `tsconfig.json` file, we now use [react-docgen-typescript](https://github.com/styleguidist/react-docgen-typescript) to parse the prop types of all React components in the current working directory. For now, only string prop values are supported. On a component-by-component basis, we use the static types if they're available, otherwise we fall back to the existing strategy via [parse-prop-types](https://github.com/diegohaz/parse-prop-types).

There aren't any automated tests for this, but I've added a TypeScript example for manual testing. I've added a Cypress test for the existing autocomplete behaviour to ensure that we haven't regressed.

In order to test this automatically, we'll need to improve our test suite to support multiple test projects, but I'm happy to leave that for another PR.


hipstersmoothie:
Works great 🎉 

screen shot 2018-12-06 at 1 02 23 am

hipstersmoothie:
One thing you might want to add is the ability to configure `react-dogen-typescript`'s config. https://github.com/styleguidist/react-docgen-typescript#parseroptions

In my storybook i use that to exclude the HTML attributes using `propFilter`.

markdalgleish:
@hipstersmoothie I might look into something like this as a follow-up PR.

DavidHooper:
Looks great, any ETA on when this will be released?

lib/getStaticTypes.js
```diff
@@ -0,0 +1,41 @@
...
+    const typesByDisplayName = keyBy(types, 'displayName');
+    const parsedTypes = mapValues(typesByDisplayName, component =>
+      mapValues(filterProps(component.props || {}), prop =>
+        parsePropTypeName(prop.type.name)
+      )
+    );
+
+    return parsedTypes;
+  } catch (err) {
+    return {};
```
jackhurley23: Should we log here just in case there's a parsing error based on a supplied tconfig?  Otherwise we're just failing silently

lib/getStaticTypes.js
```diff
@@ -0,0 +1,41 @@
...
+    const tsConfigPath = await findUp('tsconfig.json', { cwd });
+
+    if (!tsConfigPath) {
+      return {};
+    }
+
+    const { parse } = require('react-docgen-typescript').withCustomConfig(
+      tsConfigPath
+    );
+    const files = await fastGlob([`${cwd}/**/*.{ts,tsx}`, '!**/node_modules']);
```
jackhurley23: Should we be respecting the `include` and `exclude` values of the `tsconfig.json` if they exist, and then fall back to this implementation if they haven't been defined?

markdalgleish: Found this was a bit too much of a rabbit hole right now, so added an optional `typeScriptFiles` option so that you can override the default globs.

PR #36, opened 2018-12-06
feat(prettier): Format with Prettier on Cmd/Ctrl+S

jackhurley23:
Adds prettier formatting on save (CTRL/CMD + S) when on the page.

Handles multiple JSX root elements (by first wrapping the JSX in `<>{code}</>`) and maintains cursor position.


markdalgleish:
@jackhurley23 Just a heads up, I've pushed some small lint and formatting fixes.

markdalgleish:
This PR introduces this warning from webpack: https://github.com/prettier/prettier/issues/4959

Doesn't seem to cause any issues, so maybe we can just suppress the warning for now.

markdalgleish:
@jackhurley23 Just pushed a commit that suppresses the warning, with a comment explaining what it is and when we can remove it.

markdalgleish:
Awesome work. This is so cool. Gave it a trial run before merging and it works great.

PR #46, opened 2019-01-29
docs(readme): Update docs to include extra config options

pmc-a:
Update README documentation to include extra config options that can be overwritten (`port` and `openBrowser`).


markdalgleish:
Nice one! Thank you 😄

PR #48, opened 2019-02-14
Custom JSX support

grncdr:
### What

This PR makes it possible for users to easily override `React.createElement` with a function of their choice.

### Why

I'm using [Emotion](https://emotion.sh) for CSS in my project, with their Babel plugin that replaces calls to `React.createElement` with calls to Emotions own `jsx` function. Bublé has a `jsx` configuration option that allows you to accomplish the same thing, but Playroom didn't expose any way to set it.

### Design note

I opted for detecting a `createElement` global in the user provided scope, rather than exposing the Bublé option directly or adding anything in the playroom config file. I see the following benefits to this approach:

 - Makes it simple for users to pull in whatever dependencies they need to power their JSX compilation. E.g. my playroom components file has the line `export { jsx as createElement } from '@emotion/core'`.
- Relatively straightforward to implement, no need to pass extra configuration through the Playroom component tree.
- Less configuration options => less things to get wrong when setting it up.

### DIFF
--- a/src/Playroom/RenderJsx/RenderJsx.js
+++ b/src/Playroom/RenderJsx/RenderJsx.js
@@ -28,7 +28,12 @@ export default class RenderJsx extends Component {
 
     const fragment = `<React.Fragment>${jsx.trim() ||
       '<React.Fragment />'}</React.Fragment>`;
-    const { code } = transform(fragment);
+    const { code } = transform(fragment, {
+      jsx:
+        typeof scope.createElement === 'function'
+          ? 'createElement'
+          : 'React.createElement'
+    });
     const el = scopeEval(code, { ...scope, React, this: this });
 
     return el;


grncdr:
hm, I didn't expect GitHub to create a merge commit there. I will rebase and force-push my branch.

grncdr:
Hi @markdalgleish!

There's been a bunch of improvements and changes to Playroom in the last 6 months, and I'm wondering if there's any chance this patch would be reviewed if I update it? I understand everybody is busy, but I don't want to maintain a separate fork of Playroom indefinitely.

grncdr:
This is no longer needed since Playroom switched to Babel.

yenbekbay:
@grncdr I still can't find a way to specify a custom jsx pragma. Did you manage to get it working somehow without modifying the babel preset options?

grncdr:
@yenbekbay sorry no. The babel config for my project is overriding JSX globally, so I don’t use the `@jsx` pragma in my code.

PR #52, opened 2019-03-27
[WIP] Descriptions

### LINKED ISSUE #37: Prop Type Descriptions
React docgen also parses descriptions of props. It would be really cool if Playroom could display this.

Ways to use:

1. When autocompleting a prop playroom could should the description and allowed string for each prop. Here is what it looks like in VSCode

screen shot 2018-12-06 at 1 07 09 am

2. On hover you could display the same information. And even the type

screen shot 2018-12-06 at 1 10 08 am

hipstersmoothie:
closes #37

I lifted the code mostly from code mirror addon xml

- [x] Prop Type: descriptions
- [ ] Prop Type: required
- [ ] Prop Type: type
- [x] Component: description
- [ ] Test on vanilla JS

### DIFF
--- a/src/Playroom/CodeMirror-JSX.js
+++ b/src/Playroom/CodeMirror-JSX.js
@@ -1,0 +1,205 @@
+/* eslint-disable */
+
+import styles from './Playroom.less';
+
+function matches(hint, typed, matchInMiddle) {
+  if (matchInMiddle) return hint.indexOf(typed) >= 0;
+  else return hint.lastIndexOf(typed, 0) == 0;
+}
+
+function elt(tagname, cls /*, ... elts*/) {
+  var e = document.createElement(tagname);
+
+  if (cls) e.className = cls;
+
+  for (var i = 2; i < arguments.length; ++i) {
+    var elt = arguments[i];
+    if (typeof elt == 'string') elt = document.createTextNode(elt);
+    e.appendChild(elt);
+  }
+
+  return e;
+}
+
+function makeTooltip(x, y, content) {
+  var node = elt('div', styles['description-tooltip'], content);
+  node.style.left = x + 'px';
+  node.style.top = y + 'px';
+
+  document.body.appendChild(node);
+  return node;
+}
+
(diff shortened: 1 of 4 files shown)


hipstersmoothie:
@markdalgleish this should be all good now

markdalgleish:
Any chance you could add an example that exercises this? Right now, it's hard for me to verify this is working as desired.

src/Playroom/CodeMirror-JSX.js
```diff
@@ -0,0 +1,306 @@
+/* eslint-disable new-cap */
```
markdalgleish: Is this file borrowed from somewhere else? Seems like a copy-and-paste.

hipstersmoothie:
> Any chance you could add an example that exercises this? Right now, it's hard for me to verify this is working as desired.

Added descriptions and defaults to the typescript example.

src/Playroom/CodeMirror-JSX.js
hipstersmoothie: It was the addon-xml 😅 I went back for a second look and was able to use that code once it was registered to codemirror and just handle the tooltip stuff myself. 

I switched to rendering the tooltip with lit-html. I can remove that and switch back to document methods if you want, but I think the code is a little clearer this way.

hipstersmoothie: Actually I realized i could just use react and it become much clearer 👍

hipstersmoothie:
@markdalgleish Does your major re-architecture effect any of this? I'm happy to change what ever else you need me to. Really excited to get this and #55 in.

hipstersmoothie:
If these changes are too much a plugin API for suggestions would also probably fit my use case. 

As it is playroom is only useful for people that are super familiar with the components and how they work together or have the docs up in a separate tab. Even when i try to use playroom myself I feel crippled without having the type definitions helping me along.

yceballost:
would be nice to merge this feature <3

PR #55, opened 2019-03-30
Snippets

### LINKED ISSUE #38: Multiple example code
Instead of only being able to have one template page (example) it would be useful to be able to pick from multiple options

hipstersmoothie:
closes #38 

This builds off of #52 and does a little refactoring to make the snippet rendering code re-usable.

A user can define snippets in their config

```js
{
  snippets: {
    Foo: '<div>Foo</div>',
    Bar: '<div>Bar</div>',
    Baz: '<div>Baz</div>'
  }
}
```

Then when a user presses `Ctrl+Space` or `Cmd+Space` something like this will pop up:

Screen Shot 2019-03-29 at 8 05 26 PM

`Ctrl+Space` or `Cmd+Space` will also show the appropriate suggestion based on context. If used with the hints already up, it will close the hints.


michaeltaranto:
Thanks for the inspiration, we just released snippets support in [v.0.15.0](https://github.com/seek-oss/playroom/releases/tag/v0.15.0).

Let us know if you have any feedback.

Issue #45, opened 2019-01-10
Compatibility with react-pose


puregarlic:
I'm running into an issue with refs while trying to use a `react-pose` component in Playroom. I've create a reproduction repo here: https://github.com/puregarlic/playroom-pose-compat

Am I missing something? I'm not really sure why it works in a normal environment but not in Playroom.

bradbarrow:
Facing the same challenge. 

The point of difference I've narrowed down to is that **outside** of playroom, the `ref` here IS an `instanceof Element`.

and **inside** Playroom it is NOT an `instanceof Element`.

https://github.com/Popmotion/popmotion/blob/master/packages/react-pose/src/components/PoseElement/index.tsx#L169

https://github.com/Popmotion/popmotion/blob/master/packages/react-pose/src/components/PoseElement/index.tsx#L188

A potentially helpful debug log: 

![image](https://user-images.githubusercontent.com/1264276/52056350-1b12c000-25b6-11e9-9daf-31cd87ef3b93.png)

You can see that `ref` is indeed an `HTMLDivElement` but apparently that doesn't make it an instance of `Element` in this case.

Also, on a single visit to playroom...the ref get's overridden to null:

![image](https://user-images.githubusercontent.com/1264276/52056425-52816c80-25b6-11e9-9deb-f6276bcd8987.png)

And lastly a debug log from outside of Playroom:

![image](https://user-images.githubusercontent.com/1264276/52056448-675e0000-25b6-11e9-85a5-eb2377423b14.png)

I wonder if the scope-eval step in Playroom is mucking about with the refs?

I'll try to keep digging on this tomorrow

bradbarrow:
Some further observations of the lifecycle of a ref in Playroom as you edit

![image](https://user-images.githubusercontent.com/1264276/52093581-cc920f80-260f-11e9-850f-bae10b25e7ac.png)

bradbarrow:
This may be naive but I think the gist of this is that the instanceof checks are failing because the ref IS an Element but it's an instance of Element in the IFrame which is it's own DOM. 

So on the surface it seems identical but there's actually more than one DOM at play?

ryanhobsonsmith:
@bradbarrow just hit the same problem. Was able to get it working by patching like:

ryanhobsonsmith:
@bradbarrow just hit the same problem with iframe / react-pose, was able to get things working by patching react-pose and stylefire to use checks like the following:
```javascript
[...]

PR #57, opened 2019-04-05
feat: Isolate iframe code from parent window

markdalgleish:
Each iframe is now a separate document (`frame.html`), as opposed to our previous strategy where code was executed in the parent window and then rendered into blank iframes (via [react-frame-component](https://github.com/ryanseddon/react-frame-component)).

This is actually a massive re-architecture in how code is evaluated and executed, which solves a number of issues:

- Component code is now executed within the frame, so any inspection of the browser window or browser globals will now work as expected (e.g. fixes https://github.com/seek-oss/playroom/issues/45).
- Hot reloading of CSS now works, since webpack-dev-server is now inserting styles within each frame, rather than adding them to the parent document.
- CSS no longer needs to be injected into the iframe manually on page load. The iframes now own their own styles as any page typically would.

It turned out that this model is inherently slower, so I've also included the following performance improvements that seem to make up the difference:
- The error boundary wrapping each frame no longer causes a remount whenever code changes, allowing React to properly optimise each re-render when new code is provided.
- The source JSX is now only compiled once, whereas previously we were compiling it once to validate it, then once per iframe.
- The debounce for updating code is now 500ms instead of 200ms, greatly decreasing the odds of re-rendering locking up the main thread.
- Since we now have two separate HTML files being generated, I've enabled chunking to ensure that the output is better optimised to be shared between them.

## Breaking Changes

- `FrameComponent` is longer provided with the `frameWindow` and `width` props. These are redundant now that you can just access them directly as you normally would, via `window` and `window.innerWidth`.

## Development notes

I needed to update to the latest alpha release of [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin) so that it could handle our multi-page chunking strategy correctly.


lib/makeWebpackConfig.js
```diff
@@ -102,6 +105,17 @@ module.exports = async (playroomConfig, options) => {
         }
       ]
     },
+    optimization: {
+      nodeEnv: options.production ? 'production' : 'development',
+      minimize: options.production,
+      concatenateModules: options.production,
```
mattsjones: The top 3 options are actually the defaults so can probably drop unless you want to be explicit.

src/Playroom/CatchErrors/CatchErrors.js
```diff
@@ -4,33 +4,36 @@ import styles from './CatchErrors.less';
 
 export default class CatchErrors extends Component {
   static propTypes = {
+    code: PropTypes.string.isRequired,
     children: PropTypes.node.isRequired
   };
 
   state = {
     error: null,
+    erroneousCode: null,
```
mattsjones: I think `invalidCode` would be easier to type lol 🤷‍♂️

src/Playroom/Frame.js
```diff
@@ -0,0 +1,73 @@
...
+import CatchErrors from './CatchErrors/CatchErrors';
+import RenderCode from './RenderCode/RenderCode';
+
+const themesImport = require('./themes');
+const componentsImport = require('./components');
+const frameComponentImport = require('./frameComponent');
+
+const getQueryParams = (url = window.location.href) => {
+  try {
+    const hash = url.split('#')[1] || '';
```
mattsjones: Can we just use `window.location.hash` instead?

src/Playroom/Preview/Iframe.js
```diff
@@ -0,0 +1,32 @@
...
+export default class Iframe extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      loaded: false
+    };
+  }
+
+  componentDidMount() {}
```
mattsjones: Remove?

src/Playroom/Preview/Iframe.js
```diff
@@ -0,0 +1,32 @@
+import React, { Component } from 'react';
+
+export default class Iframe extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
```
mattsjones: Needs hooks 😉

markdalgleish: Yep, gonna do a sweep of this at some point.

Issue #62, opened 2019-04-16
material-ui example yields Cannot read property 'document' of undefined


mpoppins:
No doubt some simple error but if I run yarn in the root dir then in the material-ui example dir and do a yarn start I get a Cannot read property 'document' of undefined at line: 

insertionPoint: props.frameWindow.document.head

am i making a simple error?

denesreyes:
Run into the same issue, removing that line of code seems to fix the issue.

As to why it was failling I am not sure, but inspecting the props, there is no frameWindow object being passed into it. Hence the runtime error.

PR #64, opened 2019-04-17
fix(examples): Use window in favour of deprecated frameWindow prop

michaeltaranto:
As reported [here](https://github.com/seek-oss/playroom/issues/62) the material-ui and (also the reakit example) are not working since moving away from `react-frame-component` as there is no `frameWindow` provided.

Given updated iframe solution switching to `window` directly should be fine now.

Fixes https://github.com/seek-oss/playroom/issues/62

### DIFF
--- a/examples/material-ui/playroom/FrameComponent.js
+++ b/examples/material-ui/playroom/FrameComponent.js
@@ -11,7 +11,7 @@ export default class FrameComponent extends React.Component {
     this.generateClassName = createGenerateClassName();
     this.jss = create({
       ...jssPreset(),
-      insertionPoint: props.frameWindow.document.head
+      insertionPoint: window.document.head
     });
   }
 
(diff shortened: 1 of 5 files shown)


examples/material-ui/playroom/FrameComponent.js
```diff
@@ -11,7 +11,7 @@ export default class FrameComponent extends React.Component {
     this.generateClassName = createGenerateClassName();
     this.jss = create({
       ...jssPreset(),
-      insertionPoint: props.frameWindow.document.head
+      insertionPoint: window.document.head
```
markdalgleish: I think this entire line is redundant now, actually, since the code is executing within the iframe.

examples/reakit/playroom/FrameComponent.js
```diff
@@ -3,9 +3,9 @@ import { StyleSheetManager } from 'styled-components';
 import { Provider } from 'reakit';
 import theme from 'reakit-theme-default';
 
-export default ({ children, frameWindow }) => (
+export default ({ children }) => (
   <Provider theme={theme}>
-    <StyleSheetManager target={frameWindow.document.head}>
+    <StyleSheetManager target={window.document.head}>
```
markdalgleish: I don't think this is needed anymore, either.

Issue #32, opened 2018-12-03
Upgrade to codemirror2


hipstersmoothie:
Every time I install I get this 

```
warning "playroom > react-codemirror@1.0.0" has incorrect peer dependency "react@>=15.5 <16".
warning "playroom > react-codemirror@1.0.0" has incorrect peer dependency "react-dom@>=15.5 <16".
```

`react-codemirror` doesn't seem to have been published in 2 years and depends on an old version of react. `playroom` itself uses 

```json
    "react-dom": "^16.5.2",
```

So this is where the mismatch comes from. A possible solution would be to upgrade to https://www.npmjs.com/package/react-codemirror2, which allows newer versions of react

```json
    "react": ">=15.5 <=16.x",
```

siddhant1:
Will take a look at this!

hipstersmoothie:
Idk if it’s worth it m. Both react code mirror projects seem unmaintained

kaelig:
https://github.com/scniro/react-codemirror2 is great (we use it on polaris.shopify.com), way more up to date than `react-codemirror`.

hipstersmoothie:
yessssss. goodbye peer dep warnings

PR #75, opened 2019-07-23
fix(deps): Migrate to react-codemirror2

michaeltaranto:
Migrates from `react-codemirror` to `react-codemirror2`.

Also upgrading cypress and switching CI over to `yarn` so it honours the lock file.

Closes https://github.com/seek-oss/playroom/issues/32.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,17 @@
 language: node_js
 notifications:
   email: false
-before_script:
-  - npm prune
+addons:
+  apt:
+    packages:
+      # Ubuntu 16+ does not install this dependency by default, so we need to install it ourselves
+      - libgconf-2-4
+cache:
+  yarn: true
+  directories:
+    - ~/.cache/Cypress
 after_success:
-  - npm run semantic-release
+  - yarn semantic-release
 branches:
   except:
     - /^v\d+\.\d+\.\d+$/
--- a/cypress.json
+++ b/cypress.json
@@ -1,1 +1,4 @@
-{}
+{
+  "defaultCommandTimeout": 10000,
+  "video": false
+}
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -1,10 +1,17 @@
...
-  - npm prune
+addons:
+  apt:
+    packages:
+      # Ubuntu 16+ does not install this dependency by default, so we need to install it ourselves
+      - libgconf-2-4
+cache:
+  yarn: true
+  directories:
+    - ~/.cache/Cypress
```
michaeltaranto: See [cypress docs](https://docs.cypress.io/guides/guides/continuous-integration.html#Travis)

cypress.json
```diff
@@ -1 +1,4 @@
-{}
+{
+  "defaultCommandTimeout": 10000,
+  "video": false
```
michaeltaranto: Disabling video as we dont publish the assets anywhere so its a waste. Also bumping the timeout as the execution time was inconsistent.

src/Playroom/Playroom.js
```diff
@@ -352,13 +352,7 @@ export default class Playroom extends Component {
...
           style={{ bottom: this.state.height }}
         >
-          <Preview
-            code={renderCode}
-            components={components}
-            themes={themes}
-            frames={frames}
-            frameComponent={frameComponent}
-          />
+          <Preview code={renderCode} themes={themes} frames={frames} />
```
mattsjones: Why are `frameComponent` and `components` not passed anymore?

michaeltaranto: They are [unused](https://github.com/seek-oss/playroom/blob/master/src/Playroom/Preview/Preview.js#L7).

Issue #76, opened 2019-07-25
Prettier formating broken in latest release


pascalduez:
`playroom@0.9.3`

Pressing `Ctrl+s` used to trigger the Prettier formatting previously.
Now we get the following error in console:
```
TypeError: a.editorInstance.codeMirror is undefined
   handleKeyPress http://[...]/vendors~index.ea02e45bd4a3d2ddb6a2.js:13
```

PR #77, opened 2019-07-25
fix: Resolve prettier formatting error

michaeltaranto:
Fixes the prettier formatting bug introduced in the CodeMirror upgrade, plus adding a cypress test to cover the formatting scenario.

(Also upgrading lodash to get rid of the vulnerability warning)

Closes https://github.com/seek-oss/playroom/issues/76

### DIFF
--- a/cypress/integration/smokeTest.spec.js
+++ b/cypress/integration/smokeTest.spec.js
@@ -1,7 +1,9 @@
 import {
   typeCode,
   assertFrameContains,
-  assertCodePaneContains
+  assertCodePaneContains,
+  assertCodePaneLineCount,
+  formatCode
 } from '../support/utils';
 
 describe('Smoke test', () => {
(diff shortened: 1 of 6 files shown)


cypress/integration/smokeTest.spec.js
```diff
@@ -18,4 +20,11 @@ describe('Smoke test', () => {
     assertFrameContains('Foo');
     assertCodePaneContains('<Foo color="blue" />');
   });
+
+  it('formats', () => {
+    typeCode('<Foo><Foo><Bar />');
```
markdalgleish: Should these be self closing?

michaeltaranto: Nah, because when the browser types the closing brace `>` the closing tag is auto inserted after the cursor.

markdalgleish: Oh, right.

markdalgleish:
Nice work adding a test for this 👏

PR #82, opened 2019-10-28
feat: Compress code in URL

mattsjones:
Compressing the code encoded in the URL should reduce the URL length by over 50% in most cases. To allow for backwards compatibility with old URLs, the new query param is encoded as JSON. Playroom will fall back to the old format if `JSON.parse` fails.


mattsjones:
Added new `createUrl` API that is exposed via `import { createUrl } from playroom/utils`. This is also used internally to create URLs.

PR #98, opened 2019-12-10
fix: Support types from files outside of cwd

adrienharnay:
Currently, types aren't captured for this config:

```
const playroomConfig = {
  cwd: path.resolve('../my-design-system'),
};
```

This architecture being common in a monorepo, this little change does the trick.


markdalgleish:
I can't find this `absolute` option in find-up or any of its dependencies. What exactly is this option doing? From what I can see, find-up was already returning an absolute path.

adrienharnay:
And of course I made the wrong change... it is good now :)

Reference: https://github.com/mrmlnc/fast-glob#absolute

markdalgleish:
Thanks! 👍

Issue #105, opened 2019-12-17
0.11.2 breaks webpackConfig.resolve.extensions


adrienharnay:
Hello,

Since upgrading to 0.11.2, webpackConfig.resolve.extensions is not taken into account anymore.

I have components importing files without specifying the extension, and the file can either be .ts or .json, and webpackConfig.resolve.extensions is set to `['.tsx', '.ts', '.js', '.json']`.

It works fine with 0.11.1, but not with 0.11.2 (or 0.11.3).

Thanks in advance!

PR #107, opened 2019-12-19
feat: Configurable widths & themes

michaeltaranto:
This PR adds the ability to configure which themes and widths are viewable, and store them as a local user preference. In addition this preference is encoded in the url, so when sharing a link, consumers see the same setup as the author.

Also note, with the addition of a toolbar to house the preference menus, we have removed the option to left dock the editor. It already felt strange, but with these new features it really detracted from the overall experience of the preview and code taken precedence.

### Features
- Configurable themes & widths as user preference
- Configurable themes & widths in share url
- Removed left docked editor option

### Bug Fixes
- Changed order of webpack module resolution extensions array:
```diff
-['.mjs', '.js', '.json', '.ts', '.tsx']
+['.tsx', '.ts', '.jsx', '.js', '.json']
```
Should fix https://github.com/seek-oss/playroom/issues/105.

### Refactor
- Hoisted fonts, colours, shadows, etc into top level theme/variables.


src/Playroom/ToolbarItem/ToolbarItem.tsx
```diff
@@ -0,0 +1,50 @@
...
+      className={classnames(styles.root, {
+        [styles.hasSelection]: hasSelection
+      })}
+    >
+      <button
+        className={classnames(styles.button, {
+          [styles.button_isActive]: active
+        })}
+        title={title}
+        onClick={(event: MouseEvent<HTMLButtonElement>) => {
```
mattsjones: Curious, doesn't this implicitly type?

PR #114, opened 2019-12-23
fix: Align frames to centre

markdalgleish:
This ensures that frames are horizontally centred when the total width of all frames is lower than the outer viewport width. When the frames can no longer fit on the screen, the layout is exactly the same as today, seemingly left-aligned with a horizontal scrollbar.

Before:

Screen Shot 2019-12-24 at 9 47 30 am

After:

Screen Shot 2019-12-24 at 9 47 50 am

This improvement is especially noticeable when the code pane is docked to the right:

Screen Shot 2019-12-24 at 9 48 08 am

### DIFF
--- a/src/Playroom/Preview/Preview.less
+++ b/src/Playroom/Preview/Preview.less
@@ -9,10 +9,12 @@
   overflow-y: hidden;
   white-space: nowrap;
   padding: @preview-padding 0;
+  text-align: center;
 }
 
 .frameContainer {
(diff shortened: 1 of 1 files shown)


markdalgleish:
Note that I've left the frame labels left-aligned so that they're immediately visible when scrolling  through frames.  Centre alignment looks better when all frames fit on the screen, but it unfortunately causes usability issues beyond that.

PR #115, opened 2019-12-23
test: Add more integration tests

michaeltaranto:
Add more integration tests to cover url handling, visible widths toggling, copy to clipboard and assert frames are interactable.

### DIFF
--- a/cypress/integration/smokeTest.js
+++ b/cypress/integration/smokeTest.js
@@ -1,32 +1,99 @@
 import {
   formatCode,
   typeCode,
-  assertFrameContains,
+  getFirstFrame,
+  assertFirstFrameContains,
   assertCodePaneContains,
-  assertCodePaneLineCount
+  assertCodePaneLineCount,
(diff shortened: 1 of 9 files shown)


cypress/integration/smokeTest.js
```diff
@@ -1,32 +1,99 @@
...
-    ).reload();
+    cy.visit('http://localhost:9000/').reload();
+
+    typeCode('<Foo><Foo><Bar/>');
     assertCodePaneLineCount(1);
     formatCode();
     assertCodePaneLineCount(6);
   });
+
+  it('url handling - code (base64 encoding)', () => {
```
markdalgleish: So glad this is covered now! 👏

michaeltaranto: ironically, this was covered before, but the new encoding wasnt 😄

markdalgleish: lol 😅

PR #117, opened 2020-01-30
feat: Add support for inserting and previewing snippets

michaeltaranto:
Adds support for providing snippets to playroom. This allows consumers to define patterns or larger compositions of components.

Opening the snippets panel formats the code and highlights the location that the snippet will be inserted. The snippets are previewed in-situ across themes and viewports as you navigate the list, supporting a faster exploration and prototyping workflow.

Snippets can be configured as follows:
```js
// playroom.config.js

module.exports = {
  ...
  snippets: './snippets/index.js' // path to module exporting snippets
}
```

```js
// snippets/index.js

export default [
  {
    group: string, // title of snippet, eg. `Button`
    name: string,  // additional detail, eg. `Strong` 
    code: string   // <Button weight="strong">Button</Button>
  },
  ...
];
```

Closes https://github.com/seek-oss/playroom/issues/38


cypress/integration/smokeTest.js
```diff
@@ -25,75 +33,172 @@ describe('Smoke test', () => {
...
 
-  it('url handling - code (base64 encoding)', () => {
-    cy.visit(
-      'http://localhost:9000/#?code=PEZvbz48Rm9vPjxCYXIvPjwvRm9vPjwvRm9vPg'
-    ).reload();
-
-    assertFirstFrameContains('Foo');
-    assertCodePaneContains('<Foo><Foo><Bar/></Foo></Foo>');
+  it('frames are interactive', () => {
+    getFirstFrame().click('center');
```
mattsjones: Isn't this function is async? Also this test has no assertions. Is that intentional?

michaeltaranto: Just moved the test in the PR. The very act of clicking on the frame is the test. If something is blocking the frame this test fails.

README.md
```diff
@@ -89,6 +90,25 @@ To build your assets for production:
 $ npm run playroom:build
 ```
 
+## Snippets
+
+Snippets allows the definition of patterns or larger compositions of components. A snippet is previewed in-situ across themes and viewports as you navigate the list, supporting faster exploration and prototyping workflow.
```
markdalgleish:
```suggestion
Playroom allows you to quickly insert predefined snippets of code, providing live previews across themes and viewports as you navigate the list. These snippets can be configured via a `snippets` file that looks like this:
```

PR #121, opened 2020-02-05
feat: Add support for building using a baseUrl

BrunnerLivio:
## Description
This PR adds a new optional configuration option called `baseUrl`. It will be used when running `playroom build`
 
## Use Case
In case you want to prefix the playroom e.g. `https://my-website.com/playroom`.

## Other
The `baseUrl` option will get ignored when using `playroom start` - I did not see a use case for this.
If you feel like this is necessary, I'll try to make that work with the webpack dev server.

Awesome job! We love it at [Roche](https://github.com/Roche) !! :rocket:


BrunnerLivio:
Thank your, sir 🎉 @michaeltaranto

Issue #131, opened 2020-03-03
Bug: Cursor jumps to start of code editor window on error


bradbarrow:
I am trying my best to get a screen-capture of this happening.

Sometimes whilst coding in the `playroom` editor if I cause an error to be thrown (e.g. by using a variable that doesn't exist `<Button size={batm` as I type the `m` it'll throw an error (the frame will say "batm" is undefined) and my cursor jumps to the start of the editor. 

It doesn't always do this - and I'm not sure what conditions cause it. Sometimes I get errors like that and I can continue as normal, other times it jumps me around.

Posting here incase anyone else is able to re-create. (Have noted it happening on a number of engineers machines so it's not a messy local-environment thing).

michaeltaranto:
Yes we're aware of the issue. It can also occur with rapid typing. The issue is one of a few problems we are facing with codemirror at the moment.

Thanks for logging, hopefully get to this soon.

bradbarrow:
I don't know if this helps at all but I can confirm that I can reliable reproduce this by hitting backspace twice in quick succession. Other keys in quick succession don't reliably trigger this issue but a double backspace immediately sends my cursor to the top.

michaeltaranto:
Yes that is the most common occurrence for me too. Thanks for the feedback.

PR #134, opened 2020-03-10
fix: Resolve cursor position & toolbar flicker bugs

michaeltaranto:
## Bug fixes
- Fix cursor position jumping back to start
- Fix toolbar flicker and position bug when toggling snippets panel

## Refactor
- Restructured test suite to better isolate different groups of tests

Fixes https://github.com/seek-oss/playroom/issues/131


src/Playroom/CodeEditor/CodeEditor.tsx
```diff
@@ -149,7 +149,7 @@ export const CodeEditor = ({ code, onChange, previewCode, hints }: Props) => {
     return () => {
       window.removeEventListener('keydown', handleKeyDown);
     };
-  }, [dispatch, setCursorPosition]);
+  }, [dispatch]);
```
mattsjones: Looks like we should have linting for this. Should look to add this at some stage.

Issue #118, opened 2020-02-04
Props Spread and inline props don't work together


bradbarrow:
### Summary

I can use prop spreading in playroom e.g.

```
{ (props = { a: 'b', c: 'd' }) && null }

<Button {...props}>Click Me</Button>
```

I can use inline (regular) props too:

```
<Button a="b" c="d">Click Me</Button>
```

**I cannot use both at the same time**:

```
{ (props = { a: 'b', c: 'd' }) && null }

<Button {...props} e="f">Click Me</Button>
```

This doesn't work regardless of whether or not:

- you spread first, then use regular props
- you use regular props then spread
- your regular props and spread props contain the same props

### Additional challenges

Note that spreading inside an inline object also does not work:

`<Button {...({ ...props, e: 'f' })}>Click Me</Button>`

### Current Workaround

Either I don't use spreads ( :( ) or I've been doing this:

```
{ (props = { a: 'b', c: 'd' }) && null }

<Button {...(Object.assign({}, props, { e: 'f' })}>Click Me</Button>
```

Which is noticeably verbose

PR #135, opened 2020-03-11
fix: Compile with babel/standalone instead of buble

michaeltaranto:
## Bug fixes
- Allow props spreading alongside inline props

Fixes https://github.com/seek-oss/playroom/issues/118

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "@babel/preset-env": "^7.8.3",
     "@babel/preset-react": "^7.8.3",
     "@babel/preset-typescript": "^7.8.3",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -62,6 +62,7 @@
     "@babel/preset-env": "^7.8.3",
     "@babel/preset-react": "^7.8.3",
     "@babel/preset-typescript": "^7.8.3",
+    "@babel/standalone": "^7.8.7",
     "@types/base64-url": "^2.2.0",
     "@types/buble": "^0.19.2",
```
markdalgleish: Need to remove this too.

PR #136, opened 2020-03-11
feat: Support sass built-in

maraisr:
Support sass as a first class citizen. As you can see from npmtrends - sass is vastly more popular that less, and having first-class support for it would yield in less configuration for those people.

Frameworks like next.js, now [support sass built-in](https://nextjs.org/blog/next-9-3#built-in-sass-css-module-support-for-component-level-styles). 

[image](https://www.npmtrends.com/less-vs-node-sass-vs-sass-vs-less-loader-vs-sass-loader)

[Look here](https://github.com/zeit/next.js/blob/8449ebc221282d73f12a5b5608728d2efb57e8ca/packages/next/build/webpack/config/blocks/css/index.ts#L37-L61) for how next.js handles scss.

### DIFF
--- a/lib/makeWebpackConfig.js
+++ b/lib/makeWebpackConfig.js
@@ -119,6 +119,7 @@ module.exports = async (playroomConfig, options) => {
             require.resolve('style-loader'),
             cssLoader,
             postcssLoader,
+            require.resolve('resolve-url-loader'),
             require.resolve('sass-loader')
           ]
         },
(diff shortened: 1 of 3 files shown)


michaeltaranto:
Hey 👋

We made the decision to not provide any support out of the box for css preprocessors, as how projects manage styles in their components differs wildly. The usage of less is scoped to our own internal styles for playroom itself.

maraisr:
Ah! 100% my bad... Misread the line of `includedPaths` being that of playrooms source. And not consumer.

PR #137, opened 2020-03-16
fix: Support async webpack config

bradyhanna:
Allows the provided `webpackConfig` to return a promise/be an async function

```js
// playroom.config.js
module.exports = {
  // ...
  webpackConfig: async () => ({
    ...await getAsyncConfig(),
    ...otherConfig,
  })
};
```

Love the project btw 😄


markdalgleish:
Thanks for the PR! Really curious what the use case is?

bradyhanna:
> Really curious what the use case is

Our main webpack config is async to support a flow with a MFA token input

markdalgleish:
Is this something webpack itself supports in webpack config files?

markdalgleish:
So, I just figured out how to Google stuff. Looks like it can—I had no idea. Given this is a standard thing, this definitely makes sense to me now.

lib/makeWebpackConfig.js
```diff
@@ -171,7 +171,7 @@ module.exports = async (playroomConfig, options) => {
   };
 
   const theirConfig = playroomConfig.webpackConfig
-    ? playroomConfig.webpackConfig()
+    ? await Promise.resolve(playroomConfig.webpackConfig())
```
markdalgleish, line 174: Is the `Promise.resolve` call redundant here?

maraisr: I think maybe it is. Afaik you cant await something that isnt thenable? Or at least I've got eslint rules that complain when you're using await on non-thenables.

bradyhanna: It doesn't look like `webpack-merge` supports promises https://github.com/survivejs/webpack-merge/issues/81

Issue #125, opened 2020-02-13
Code can break out of the iframe and modify parent page (XSS)


tom-sherman:
It is currently possible to access `window.parent` inside of the render iframes which allows a malicious actor to take advantage of XSS or some accidentally dodgy code to crash the Playroom environment.

Is it within the scope of this project to guard against these sorts of problems?

## Example
```javascript
<div>
	{window.parent.document.body.innerHTML = '<h1>You\'ve been hacked lol</h1>'}
</div>
````
https://seek-oss.github.io/braid-design-system/playroom/#?code=N4Igxg9gJgpiBcIA8AhCAPAfAHQHbAHcBLXKCAgOgAcBDAJxlwBcKywBXAW0ZYCNoAnhRK4YdABIAVALIAZAAQBeeQHIkACwCMmAJoR22bCoBuMebxiN56mmADWMKPIA2EZ0gD0WzCoC+ntCwQXyA

## Potential fixes

Adding `sandbox="allow-scripts"` to the iframe component fixes this issue at the cost of completely breaking HMR.

I tried `eval`ing the code in a worker but I don't think React components are structured-clonable

Maybe it's possible to use realms-shim with a bit of work?

mattcompiles:
This should be resolved by #157

PR #157, opened 2020-05-21
feat: Add iframeSandbox config

parshap:
Hi! We’re looking at integrating Playroom on our design system style guide at [NerdWallet](https://www.nerdwallet.com/). Given Playroom’s inherent ability to execute arbitrary code, one of our security engineers had some recommendations:

1. Move the site to its own domain where cookies for other NerdWallet applications will not be available
1. Use the `sandbox` attribute in Playroom’s iframe to sandbox arbitrary code execution

This PR is about the second item and implements the `sandbox` attribute in the `<Iframe>` component driven by `playroom.config.js` config. Credit for the implementation also goes to @mhensley-nw.

This security concern was also raised in #125.

### Unbreaking Webpack HMR

Like #125 mentions, using `sandbox=”allow-scripts”` breaks Webpack hot module replacement in development. This is because Webpack expects an `Origin` header and `sandbox` makes requests omit `Origin`. This is fixed by using webpack-dev-server’s `disableHostCheck` option, which makes Webpack not expect the `Origin` header. More information in this Webpack issue: https://github.com/webpack/webpack-dev-server/issues/1604.

### Configuration

This is an opt-in feature using the `iframeSandbox` key in `playroom.config.js`. A value of `true` defaults to `”allow-scripts”` or a custom string can be used. We think `”allow-scripts”` might be a sane default, so you may consider making that the default in the future.

### DIFF
--- a/cypress/projects/basic/playroom.config.js
+++ b/cypress/projects/basic/playroom.config.js
@@ -3,4 +3,5 @@ module.exports = {
   snippets: './snippets',
   outputPath: './dist',
   openBrowser: false,
+  iframeSandbox: true,
 };
--- a/cypress/projects/basic/playroom.config.js
+++ b/cypress/projects/basic/playroom.config.js
@@ -3,5 +3,4 @@ module.exports = {
   snippets: './snippets',
   outputPath: './dist',
   openBrowser: false,
-  iframeSandbox: true,
 };
--- a/examples/seek-style-guide/playroom.config.js
+++ b/examples/seek-style-guide/playroom.config.js
@@ -47,5 +47,5 @@ module.exports = {
         ],
       },
     }),
-  iframeSandbox: true,
+  iframeSandbox: 'allow-scripts',
 };
--- a/examples/seek-style-guide/playroom.config.js
+++ b/examples/seek-style-guide/playroom.config.js
@@ -47,4 +47,5 @@ module.exports = {
         ],
       },
     }),
+  iframeSandbox: true,
 };
--- a/src/Playroom/Frames/Iframe.tsx
+++ b/src/Playroom/Frames/Iframe.tsx
@@ -15,11 +15,11 @@ interface IframeProps extends AllHTMLAttributes<HTMLIFrameElement> {
 const playroomConfig = (window.__playroomConfig__ = __PLAYROOM_GLOBAL__CONFIG__);
 
 export default function Iframe({
-                                 intersectionRootRef,
-                                 style,
-                                 src,
-                                 ...restProps
-                               }: IframeProps) {
+  intersectionRootRef,
+  style,
+  src,
+  ...restProps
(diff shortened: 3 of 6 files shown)


src/Playroom/Frames/Iframe.tsx
```diff
@@ -12,6 +12,18 @@ interface IframeProps extends AllHTMLAttributes<HTMLIFrameElement> {
   intersectionRootRef: MutableRefObject<Element | null>;
 }
 
+const playroomConfig = (window.__playroomConfig__ = __PLAYROOM_GLOBAL__CONFIG__);
```
parshap, line 15: I wasn't sure if I should read the config in this component or if the value should be read elsewhere and passed as a prop. I tried to follow existing patterns, which looked like it's ok for components to read config.

cypress/projects/basic/playroom.config.js
```diff
@@ -3,4 +3,5 @@ module.exports = {
   snippets: './snippets',
   outputPath: './dist',
   openBrowser: false,
+  iframeSandbox: true,
```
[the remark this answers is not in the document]
parshap: Turns out this was breaking Cypress (makes sense because `sandbox` limits iframe access); so I moved it to `examples/seek-style-guide.playroom.config.js`.

mattcompiles:
Hey Mate, Thanks for the PR. Controlling the sandbox attribute seems reasonable. However, I'm not sure about the `iframeSandbox: true` behaviour. I think it's not clear how secure the iframe now is. Maybe it'd be cleaner to just leave the string forward behaviour and let users decide what value they pass? 

Probably also worth mentioning that at least `allow-scripts` is required for playroom to work?

parshap:
Hey @mattcompiles. I think you're right that a default of `allow-scripts` may not be right for all use cases and doesn't make clear what's happening. I made your suggestion change of requiring an explicit string value and adding a note about needing `allow-scripts` in the readme. Thanks for the input!

examples/seek-style-guide/playroom.config.js
```diff
@@ -47,4 +47,5 @@ module.exports = {
         ],
       },
     }),
+  iframeSandbox: true,
```
mattcompiles, line 50: Need to update this example

parshap: Whoops, thanks.

```suggestion
  iframeSandbox: 'allow-scripts',
```

penx:
@parshap @mattcompiles it would be great to note these security concerns/recommendations in the readme, especially putting Playroom on to its own domain. From my understanding, hosting Playroom on a github.io subdomain would be secure (as far as it couldn't read any sensitive cookies/localstorage), but I expect people may want to deploy Playroom in other ways and may not be immediately aware of these considerations.

PR #161, opened 2020-05-29
feat: Add paramType config to use standard search queries

mattcompiles:
This is useful when creating private playrooms that sit behind auth redirects. In these cases hash based queries will be stripped from the URL.


maraisr:
Can you give us a bit more detail without having to read through the code @mattcompiles?

PR #170, opened 2020-06-25
fix: Preserve extra params in URL

mattcompiles:
This change ensures only the `code` param in the URL is modified. This allows custom params to be retained. 

Note: A lockfile refresh needed to be done for a babel issue unrelated to this change.

### DIFF
--- a/cypress/support/utils.js
+++ b/cypress/support/utils.js
@@ -52,6 +52,7 @@ export const toggleSnippets = () =>
   cy.get('[data-testid="toggleSnippets"]').click();
 
 export const filterSnippets = (search) =>
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
   cy
     .get('[data-testid="filterSnippets"]')
     .type(search, { force: true })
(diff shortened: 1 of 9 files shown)


cypress/support/utils.js
```diff
@@ -52,6 +52,7 @@ export const toggleSnippets = () =>
   cy.get('[data-testid="toggleSnippets"]').click();
 
 export const filterSnippets = (search) =>
+  // eslint-disable-next-line cypress/no-unnecessary-waiting
```
mattcompiles, line 55: Due to lockfile update adding new cypress lint rules. Ignoring for now.

cypress/support/utils.js
```diff
@@ -74,6 +75,7 @@ export const assertSnippetCount = (count) =>
 
 export const assertFirstFrameContains = (text) => {
   getFirstFrame().then(($el) =>
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
```
mattcompiles, line 78: As above

src/StoreContext/StoreContext.tsx
```diff
@@ -417,7 +420,7 @@ export const StoreProvider = ({
           themes: parsedThemes,
           widths: parsedWidths,
         } = JSON.parse(
-          lzString.decompressFromEncodedURIComponent(String(params.code))
+          lzString.decompressFromEncodedURIComponent(String(params.code)) ?? ''
```
mattcompiles, line 423: Lockfile related change. Type signature was updated.

PR #181, opened 2020-08-14
fix: Add theme to compressParams

MarijnBrosens:
Add theme to compressParams so PreviewUrl works with multibranding.


mattcompiles:
Thanks for the PR bud. Noticed this issue the other day but hadn't tracked down the source. Could you just update the type declaration file to match as well?

mattcompiles:
Thanks for the PR

Issue #40, opened 2018-12-06
Modify Head


hipstersmoothie:
It would be nice to have a way to include things in the head of the page. 

For example here i need to include a company font icon hosted on an internal CDN. Something like storybook's `preview-head.html` would be perfect

malcolm-kee:
There is a workaround in the meantime, by "abusing" framecomponent to append scripts/stylesheet in the head.

Example:

```tsx
import * as React from 'react';

function useStylesheet(url: string) {
  React.useEffect(() => {
    const link = document.createElement('link');
    link.href = url;
    link.rel = 'stylesheet';
    const head = document.getElementsByTagName('head')[0];
    head.appendChild(link);
    return function() {
      head.removeChild(link);
    };
  }, [url]);
}

export default function PlayroomFrame(props: { children: React.ReactNode }) {
  useStylesheet('https://mycdn/mystyle.css');

  return <>{props.children}</>;
}
```

PR #185, opened 2020-09-11
feat: Define Components at beginning and end of body

alexbchr:
Adds 2 options to `playroom.config.js`, which are `bodyStart` and `bodyEnd`. They work similarly to `frameComponent`, which means that they receive a file path to a component which will be rendered at the beginning/end of body (well in fact in the root div).

Example usage:
```tsx
// playroom.config.js
{
  // ...
  bodyEnd: './bodyEnd.js'
  // ...
}

// bodyEnd.js
import React from 'react'

export default () => (
  <script type="text/javascript">
    {/* Put Analytics script here */}
  </script>
)
```

This was necessary for my team as we wanted to plug an Analytics library on our playroom, which of course requires us to add a script in the `<body />`. It could also be used to provide a watermark, link or anything absolutely positioned on the page, maybe for branding purposes. Furthermore, I think this could also fix issue #40, as using the new `bodyStart` option and a library such as [React Helmet](https://github.com/nfl/react-helmet), this use case could be covered.

I am also aware that some things are missing from this PR such as examples and documentation, which I will include once I know this is relevant to this repo. 

Also, I am considering adding a similar feature to the preview page, so we can have analytics there too. Prop names could be something like `previewBodyStart` and `previewBodyEnd`.

### DIFF
--- a/lib/defaultModules/bodyEnd.js
+++ b/lib/defaultModules/bodyEnd.js
@@ -1,0 +1,1 @@
+export default () => null;
--- a/lib/defaultModules/bodyStart.js
+++ b/lib/defaultModules/bodyStart.js
@@ -1,0 +1,1 @@
+export default () => null;
--- a/lib/makeWebpackConfig.js
+++ b/lib/makeWebpackConfig.js
@@ -52,6 +52,12 @@ module.exports = async (playroomConfig, options) => {
         __PLAYROOM_ALIAS__FRAME_COMPONENT__: playroomConfig.frameComponent
           ? relativeResolve(playroomConfig.frameComponent)
           : require.resolve('./defaultModules/FrameComponent'),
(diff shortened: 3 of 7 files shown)


michaeltaranto:
Hey 👋 

Just having a look at your approach, is there a need to render this from inside the Playroom app? Or could this be achieved by opening up config to provide your own html template for the built app?

maraisr:
I'd second that @michaeltaranto, my subjective opinion is that "before/after(render)" _hooks_ seem bug prone, and probably won't suite all use-cases. 

Much easier if it was just a blanket override of the template; such that yeah as you've said @michaeltaranto, simply overriding the template.html. And just articulate in the docs or something which element the app mounts into.

alexbchr:
That's a great idea, I will try to look at this when I have some time this week.

Should I close this PR and open a new one once the code is ready?

alexbchr:
Closing in favor of #189

PR #187, opened 2020-09-14
fix(Preview): Resolve SplashScreen rendering behind content

michaeltaranto:
Isolates the rendered content in Preview mode to its own stacking context to ensure it's below the `SplashScreen`.

### DIFF
--- a/src/Playroom/Preview.less
+++ b/src/Playroom/Preview.less
@@ -1,0 +1,4 @@
+.container {
+  position: relative;
+  z-index: -1;
+}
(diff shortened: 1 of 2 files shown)


mattcompiles (approved):

crswll:
This PR is causing a few issues for me. My frame is like:

```jsx
import React from 'react'

export default ({ theme, children }) => (
  <>
    <div className={`min-h-screen space-y-6 font-default antialiased bg-canvas pb-8 ${theme}`}>
      {children}
    </div>
  </>
)
```
and the preview ends up looking like:
```
<div class="min-h-screen space-y-6 font-default antialiased bg-canvas pb-8 theme-name">
  <div class="Preview__container--23y7x">
    ... actual code from preview...
  </div>
  <div class="SplashScreen__root--3KpNS SplashScreen__hideSplash--9Cnrv">
    ...
  </div>
</div>
```
which breaks a lot. I also can't see the preview at all without removing the negative z-index.

PR #188, opened 2020-09-14
Correctly display widths specified in non-px units

jesstelford:
Specifying a width with any unit other than `px` results in the following:

Screen Shot 2020-09-14 at 1 22 13 pm

ie; it always appends `px`.

This change uses the assumption that when the width value ends in a number, it is an implicit `px` value, otherwise, it leaves it alone and uses the exact value supplied for display.

### DIFF
--- a/src/Playroom/Frames/Frames.tsx
+++ b/src/Playroom/Frames/Frames.tsx
@@ -21,7 +21,11 @@ export default function Frames({ code, themes, widths }: FramesProps) {
   const scrollingPanelRef = useRef<HTMLDivElement | null>(null);
 
   const frames = flatMap(widths, (width) =>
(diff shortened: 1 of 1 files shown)


michaeltaranto:
Hey mate, typescript isn't happy passing a `number` to the regex `test` function, just need to parse to a string before testing it

jesstelford:
🤦  I couldn't replicate the failure locally and had no idea what was going wrong... Turns out I was still on the `master` branch 🤦

jesstelford:
Thank you for the pointer. Should be fixed now 🤞

michaeltaranto (approved):

michaeltaranto:
Thank you sir 🙇

PR #196, opened 2020-09-25
docs: Add Mesh example to README

lauriejones:
We've got a themed example of playroom running on our design system, Mesh, that we'd like to add. 🙂


markdalgleish (approved):
Thank you!! Nice work 👏

Issue #197, opened 2020-09-28
preview z-index issue


riencoertjens:
I ran into this issue after upgrading our package, this pr: #187 is the culprit

making the z-index of the container negative puts it behind the <body> element and not clickable, or in case you give the body a background color it will just show a blank page

in [this example](https://seek-oss.github.io/braid-design-system/playroom/preview/#?code=N4Igxg9gJgpiBcIA8BBANjATgFwATYgDsYBeAHRAEtCAzCCgPiQBUYAPbBgdRjUgFsY+CLgAKaAIYBPTBAj8AhEgD0rDk2XosnMoSQAhAK7YChBthgBnbCqMmiDEABoQ2ABYxBCEBIAOEsBAAXyA) you can't click the button because of this

PR #198, opened 2020-09-28
fix(Preview): Don't set negative z-index for preview content

riencoertjens:
fixes change made in #187

rather than just reporting, this should fix issue #197

simply setting `position: relative;` on the container of the preview content will prevent child elements within the container to stack above the `SplashScreen` because they will be stacked relatively to the container (and not it's parent)

### DIFF
--- a/src/Playroom/Preview.less
+++ b/src/Playroom/Preview.less
@@ -1,4 +1,3 @@
 .container {
   position: relative;
-  z-index: -1;
 }


crswll:
The wrapper introduced in #187 also has some interesting effects on the Frame: https://github.com/seek-oss/playroom/pull/187#issuecomment-700165520

markdalgleish:
Thanks for this PR! I ended up fixing it in a slightly different way, but I still really appreciate this.

PR #200, opened 2020-10-03
feat: Supports react-dom concurrent mode for frame renders

maraisr:
TL;DR: 
- need to support concurrent mode
- thus need frame render to mount using `createRoot`.
- can do through a simple boolean prop (as this PR is doing)
- can do through config function `frameRenderFunction(Frame): void`
- can breaking change, and make the frame do its own rendering (like [sku does](https://github.com/seek-oss/sku/blob/master/entry/client/index.js#L14-L23))

---

This PR aims to allow Playroom to support Concurrent mode (optionally). Our design system is now required to support concurrent mode (yes we've jumped the gun early). As such we'd like Playroom to mount frames with the new `createRoot` api. Granted this is all experimental things, and will likely need an update as soon as it drops.

This is the less impactful change. One thing that came to mind, as an alternative implementation would be to do something like [sku does](https://github.com/seek-oss/sku/blob/master/entry/client/index.js#L14-L23) and simply make the consumers `frame` javascript file return a function, that is in charge of rendering how they please—rather than just return a component.

I guess to counter that last comment, you could instead introduce a playroom config method that does this, aka `frameRenderFunction(Frame): void`.

Please let me know of a direction.


package.json
```diff
@@ -23,7 +23,7 @@
     "preview:build": "cd examples/braid-design-system && yarn && yarn build",
     "commit": "git-cz",
     "lint": "eslint . && prettier --list-different '**/*.{js,md,less,ts,tsx}' && tsc --noEmit",
-    "format": "prettier --write '**/*.{js,md,less,ts,tsx}'",
+    "format": "prettier --write \"**/*.{js,md,less,ts,tsx}\"",
```
maraisr, line 26: This is done to support windows.

examples/concurrent-mode/playroom.config.js
```diff
@@ -0,0 +1,99 @@
...
+  `,
+  webpackConfig: () => ({
+    resolve: {
+      alias: {
+        // These are done, because in example playroom _root_ imports them in its context, so it doesnt pick up the examples node_modules
+        react: require.resolve('react'),
+        'react-dom': require.resolve('react-dom'),
+      },
+    },
+    module: {
```
maraisr, line 43: this was all a copy-paste job of braid example. Let me know if you'd rather not. (braid is probably already concurrent mode compat, but if not i can use something else)

maraisr:
Friendly ping @markdalgleish @michaeltaranto @mattcompiles

PR #201, opened 2020-10-12
feat: Add 'scope' option for providing custom scope variables

markdalgleish:


michaeltaranto (approved):

README.md
```diff
@@ -124,11 +125,24 @@ If your components need to be nested within custom provider components, you can
...
+## Custom Scope
+
+You can provide extra variables within the scope of your JSX via the `scope` option, which is a path to a file that exports a `useScope` Hook that returns a scope object. For example, if you wanted to expose a context-based `theme` variable to consumers of your Playroom:
+
+```js
+import { useTheme } from '../path/to/your/theming-system';
+
+export default function useScope() {
+  return {
+    theme: useTheme(),
```
maraisr, line 144: @markdalgleish  this is great! 

How does this get used? like is it a _magical_ object that is accessible in the code pane? Ie can we `scope.XYZ` and use that? Or what is a usecase for this?

PR #204, opened 2020-11-18
fix(config): Add .mjs as an extension for Webpack to resolve

nwalters512:
https://github.com/seek-oss/playroom/pull/107 dropped `.mjs` from the list of extensions Webpack will resolve. This is causing problems for us - `graphql` is in the import path of some of our components, and https://github.com/graphql/graphql-js/issues/1272 is biting us. This cannot be fixed in userspace via `webapckConfig` in `playroom.config.js`, since the `.mjs` extension needs to appear before the `.js` extension, but the user-provided `resolve.extensions` is concatenated to the *end* of Playroom's base Webpack config.

An alternative solution would be to customize the way that `webpack-merge` merges `resolve.extensions` so that user-provided extensions always appear first. I'd be happy to pivot this PR to that if that's what the maintainers would prefer.


maraisr:
`mjs` actually require a bit more than just resolution. You'd need to couple this with a webpack module rule:

```js
config.module.rules.push({
	test: /\.mjs$/,
	include: /node_modules/,
	type: 'javascript/auto',
});
```

both of this you could just do in consume land with the webpack override.

nwalters512:
Ah, interesting! Thanks for the tip. Before making this PR, I had tested locally by editing Playroom files in `node_modules`, and the build succeeded just with the addition of the `.mjs` extension. When I tested with Webpack config overrides, I didn't add the module rule. After adding that, things seem to work. Here's a simplified version of what my Playroom config looks like now:

```js
// playroom.config.js
module.exports = {
  webpackConfig: () => ({
    resolve: {
      extensions: ['.mjs'],
    },
    module: {
      rules: [
        {
          test: /\.mjs$/,
          include: /node_modules/,
          type: 'javascript/auto',
        },
      ],
    },
  }),
};
```

Would be happy to add the module rule to this PR so that things "just work" for other folks.

maraisr:
It all stems from Webpack not handling mjs properly. Or properly, who knows. The extension is fractured and nobody really knows! But this plugin resolved all: 
https://github.com/lukeed/webpack-modules

michaeltaranto (approved):

PR #207, opened 2020-12-15
Add portfinder to start command

ryanhatfield:
## Problem:

After updating my local PHP version, an executable called `cslistener` started using port 9000. 
The `playroom start` command succeeded, and gave me no indication there was an issue with the port already being in use.

## Solution:

This PR adds the `portfinder` library to automatically find the first available port, starting with the port from the config file.


michaeltaranto (approved):
Thank you!

PR #209, opened 2021-01-10
fix: Upgrade to TypeScript 4

etaoins:
Since seek-oss/sku#535 it has been pulling in both TypeScript 4 (for itself) and TypeScript 3 (for Playroom). This is wasteful and in some situations Yarn will pick the `tsc` from TypeScript 3.

This bumps our `typescript` and `eslint-config-seek` deps to require TypeScript 4.


jahredhope (approved):

markdalgleish (approved):

markdalgleish:
Thanks for the PR 👍

Issue #156, opened 2020-05-19
Is this for React component libraries only?


chris-dura:
It'd be really great to get this working with native Web Components. Or even, in particular, StencilJS uses TypeScript and is JSX "enabled", at least on some level. Stencil can also output to React components.

Is it possible that this project could be made to support components written in other frameworks, or other JSX-enabled syntaxes?

If it is possible, at a high-level, what would we have to do to make a compatible input for Playroom?

BrunnerLivio:
We use StencilJS + React integration with Playroom. Unfortunately, I can't share the repository.

Basically what you need to do is build the library with the Stencil compiler and reference your destination folder as described in the documentation.
In our case this looks something like this:


```javascript
module.exports = {
  // I use an absolute path because the library is part of the node_modules folder.
  // In your case most probably its a relative part.
  components: "@design-system/react",
  themes: "@design-system/react/dist-transpiled/themes",
  ...
```

Theming is a more tricky one. The way it works with our design system is the "Ionic"-way. Basically you need to set a `mode` attribute to e.g. the root body element

```
<body mode="dark">
...
```

In order to integrate it with the Playroom, I simply created a `src/themes/index.ts` file which contains the following code. This file needs to be configured as well in the Playroom config (see above). Basically this file will provide all the possible theming names.

```javascript
export const deviceTheme = "device";
export const lightTheme = "light";
export const darkTheme = "dark";
```


and the `FrameComponent.js` of Playroom can then load the theme via the given strings.

```javascript
import React from "react";

import "style-loader!css-loader!@design-system/core/dist/web/web.css";

export default ({ children, theme }) => {
  return <div mode={theme}>{children}</div>;
};
```

![Preview](https://user-images.githubusercontent.com/9899423/82360051-13746600-9a09-11ea-9a1a-955bfd2311da.png)

I hope this helps? :)

chris-dura:
@BrunnerLivio -- Thanks! I wasn't expecting an answer so quickly, so it might be a minute before I can try this out, but a few questions to clarify --

```json
// playroom.config.js

module.exports = {
  components: "@design-system/react",
...
```
1. I've never output anything from Stencil other than the native web components, so I assume that `@design-system/react` is what's generated by using the `react-output-target` that Stencil has in their documentation?
https://stenciljs.com/docs/react#stencil-config-setup

[...]

PR #221, opened 2021-05-08
WIP: Support web components (or any components)

michaelwarren1106:
IMO, there are two things stopping any kind of component from being able to be easily used with Playroom.

- importing the component code into the Playroom build
- Getting hints registered in the proper format

This PR provides an example for TS/ES web components (Lit) to be imported through the standard `components.js` but bypassing the normal export option by just exporting `{}`

Also, I added a new config option in playroom.config.js `componentHints` that takes either a function or a hints object. The results of that config property are passed to the CodeEditor component (manual hints are preferred if they exist) so that hints can be manually generated and passed into playroom in a case where the component authoring mechanism doesnt allow attribute types, props, defaults to be automatically deciphered.

This is my first contribution to an OSS project, especially one of this size, so please tell me where I've gone wrong or if you'd like to see a different approach. I wasn't sure that running cypress tests was relevant since i've only added a config prop, but I'm open to adding features to test the new config if need be.

This PR is something similar to work that my coworker has done locally to an install of playroom we have running. We use Lit web components, and having this config prop be native would simplify our playroom considerably, and let us just parse output from analyzer tools such as runem's [web-component-analyzer](https://github.com/runem/web-component-analyzer) or the [new one open-wc is working on](https://github.com/open-wc/custom-elements-manifest/tree/master/packages/custom-elements-json-analyzer)

addresses #156


michaelwarren1106:
Bumping this PR just to get the initial build running. Is there anything else I need to do to at least get a build?

PR #237, opened 2021-09-13
fix: Correct rule.use to an array

mihkeleidast:
According to [webpack documentation](https://webpack.js.org/configuration/module/#ruleuse), `rule.use` should be an array. I believe it was once allowed to be an object and is still backwards compatible.

For more background, I use [external-svg-sprite-loader](https://github.com/bensampaio/external-svg-sprite-loader) which iterates over the rules/loaders and expects `rule.use` to be an array per the webpack docs. Trying to use playroom but running into bensampaio/external-svg-sprite-loader#58 at the moment, this change should fix the issue.


michaeltaranto (approved):
Thanks!

PR #238, opened 2021-09-13
feat: Update classnames

mihkeleidast:
This fixes a deprecation warning new users currently get when installing playroom:
```
npm WARN deprecated @types/classnames@2.3.1: This is a stub types definition. classnames provides its own type definitions, so you do not need this installed.
```


michaeltaranto (approved):
Thank you 👏

PR #240, opened 2021-09-13
feat: Upgrade webpack-dev-server

mihkeleidast:
Upgrades webpack-dev-server to latest in order to resolve two npm deprecation warnings when installing playroom:
```
npm WARN deprecated chokidar@2.1.8: Chokidar 2 will break on node v14+. Upgrade to chokidar 3 with 15x less dependencies.
npm WARN deprecated fsevents@1.2.13: fsevents 1 will break on node v14+ and could be using insecure binaries. Upgrade to fsevents 2.
```

Not sure if there are any actual issues on Node 14 when using older webpack-dev-server, but I like to be better safe than sorry and not introduce issues into our codebase.

Last one from me for today, ran into these four issues when installing :)


michaeltaranto (approved):
Good stuff! Thanks for being so thorough 👏

PR #248, opened 2021-10-04
fix: prevent exception on start when no TS config is found

### LINKED ISSUE #247: Version 0.27.7 throws when no TS config file is found
Just upgraded playroom from `0.27.2` to `0.27.7` and I now get an exception when starting it.
Note: the project is not using TS.  

Probably introduced in #242 

```
(node:31436) UnhandledPromiseRejectionWarning: TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined                                                                   
    at validateString (internal/validators.js:124:11)                                                    
    at Object.dirname (path.js:1248:5)                                                                   
    at module.exports ([...]/playroom/node_modules/playroom/lib/getStaticTypes.js:27:25)                                                                                           
    at async module.exports ([...]/playroom/node_modules/playroom/lib/makeWebpackConfig.js:19:23)                                                                                  
    at async module.exports ([...]/playroom/node_modules/playroom/lib/start.js:7:25)                                                                                               
(Use `node --trace-warnings ...` to show where the warning was created)           
```

pascalduez:
This PR should fix #247 

`undefined` is passed to `path.dirname`, on projects with no TS config file.

### DIFF
--- a/lib/getStaticTypes.js
+++ b/lib/getStaticTypes.js
@@ -24,7 +24,6 @@ module.exports = async (playroomConfig) => {
   } = playroomConfig;
 
   const tsConfigPath = await findUp('tsconfig.json', { cwd });
-  const basePath = path.dirname(tsConfigPath);
 
   if (!tsConfigPath) {
     return {};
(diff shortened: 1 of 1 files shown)


artmsilva:
Bumping this. We would like to use the latest version however are blocked due to this bug.

michaeltaranto (approved):

michaeltaranto:
Apologies for taking so long to look at it. Thank for this!

Issue #66, opened 2019-04-26
Support state


markdalgleish:
It'd be great if users could make their designs stateful.

My thinking is that we could do this in a code-oriented way by allowing usage of React hooks in the following way:

```
const [ firstName, setFirstName ] = useState('');
const [ lastName, setLastName ] = useState('');

<TextField label="First name" value={name} onChange=(event => setFirstName(event.target.value)} />
<TextField label="Last name" value={name} onChange=(event => setLastName(event.target.value)} />
```

For this to work, we'd need to treat the last *n* number of JSX elements as an implicitly returned fragment, i.e. behind the scenes we'd turn it into a component that looks like this:

```
() => {
  const [ firstName, setFirstName ] = useState('');
  const [ lastName, setLastName ] = useState('');

  return (
    <React.Fragment>
      <TextField label="First name" value={name} onChange=(event => setFirstName(event.target.value)} />
      <TextField label="Last name" value={name} onChange=(event => setName(event.target.value)} />
    </React.Fragment>
  );
};
```

kvendrik:
As an alternative to consider last week we did some thinking on if we would be able to let people set hooks within the JSX somehow using say a custom component. Not sure if this should be something that is directly implemented into Playroom itself but thought the idea was interesting. I do wonder though if and how limiting this would be.

```tsx
<PlayroomState defaultState={false}>
  {(toggled, setToggled) => {
    <Button onClick={() => setToggled(true)}>Open</Button>
    <Modal
      open={toggled}
      onClose={() => setToggled(false)}
    >
      Modal content
    </Modal>
  }}
</PlayroomState>
```

danrosenthal:
Here's a quick-n-dirty pull request that creates some wrapped functions and uses context magic to enable no-config state that gets us pretty close to what I consider to be the ideal prototyping API from a designer's perspective.

https://github.com/Shopify/polaris-react/pull/1372

I see an approach like this as following the 80/20 rule. Any components that require state, or any components that require lots of config, would follow this approach. That gets you to about 80% of your rapid prototyping needs. Having **hooks** available gets you the other 20% of the way there.

I see this as the ideal compromise for designer prototyping productivity.

markdalgleish:
That's interesting. Anything component-based is inherently supported by Playroom already. I see the potential here for people to create component-oriented state management libraries designed for prototyping. If you come up with something along these lines, I might see if I can find a way to talk about it in the readme.

hipstersmoothie:
[...]

PR #262, opened 2022-09-15
feat(config): Add .processCode config as a hook for complex usecases

jesstelford:


jesstelford:
For a slightly more complex example of what `processCode` enables, we're now using this to handle the example @markdalgleish outlined in the OP of #66 :

```typescript
export default function processCode(code: string) {
  // Everything after the last blank link is considered "rendered"
  const codeLines = code.trim().split('\n\n');
  const jsx = codeLines.pop() || '';

  // Wrap the entire code block in an IIFE so we have a scope where
  // arbitrary JS (including React state) can be executed
  return `{
      (() => {
        ${codeLines.join('\n\n')}
        ${
          // When there's already a `return`, leave it as-is
          jsx.trim()?.startsWith('return')
            ? jsx
            // Otherwise, wrap it in a return + fragment for safe rendering.
            : `return (
            <>
              ${jsx}
            </>
          )`
        }
      })()
  }`;
}
```

jesstelford:
Some investigation reveals that code formatting does not consider any changes returned from `processCode`. In the OP example, the code is formatted oddly, because prettier doesn't understand that it's wrapped in the IIFE.

If we were to call `processCode` before the formatter, we'd end up inserting the IIFE into the Playroom editor, which is not what we want.

For our particular usecase, we've disabled formatting by patching our installation of Playroom, but that's not a great solution.

I'm not sure if this issue should block the PR or not?

alex-page:
@playroom-maintainers any interest in this contribution?

alex-page:
@seek-oss/playroom-maintainers any interest in this contribution?

markdalgleish:
We're definitely keen on exploring this! I do think the code formatting issue is a show-stopper, so I don't think we should open up such a low level feature if it won't integrate cleanly with the rest of Playroom. It might be worth exploring more high-level, built-in support for multiple expressions beyond just JSX.

I can see a couple of viable options for this:
- We could use syntax similar to what you showed where you can write multiple expressions with an implicit return of JSX at the end.

  I can see this requiring some AST transformations to handle the implicit return — splitting on multiple line breaks won't work because this is valid within your JSX.

  Managing the formatting of it could be tricky too since it's not valid code as it's written. We already have to wrap everything in a fragment behind the scenes to make formatting of sibling JSX elements work, and then remove the fragment from the formatted code afterwards. This would need to be an even more complicated version of that.

  I do have some concerns with this approach since it's no longer "just JSX" and instead becomes a bit of a weird hybrid. It might feel a bit unclear what the syntax rules are. Maybe it's ok.
- We could keep the JSX handling as it is now, but add an additional opt-in pane to the UI that contains the setup code where you can put your hooks etc. The challenge here will be maintaining Playroom's simplicity, which to me is a big part of what sets it apart.

PR #264, opened 2022-11-03
Add snapshot workflow

askoufis:
Adding a snapshot workflow so I can test a branch on Braid's playroom.

### DIFF
--- a/.github/workflows/snapshot.yml
+++ b/.github/workflows/snapshot.yml
@@ -1,0 +1,31 @@
+name: Snapshot
+
+on: workflow_dispatch
+
+jobs:
+  release:
+    name: Publish snapshot version
+    runs-on: ubuntu-latest
+    env:
+      CI: true
+    steps:
+      - name: Check out repo
+        uses: actions/checkout@v3
+        with:
+          # This makes Actions fetch all Git history for semantic-release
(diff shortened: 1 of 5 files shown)


.github/workflows/snapshot.yml
```diff
@@ -0,0 +1,31 @@
...
+          node-version-file: 'package.json'
+
+      - name: Install dependencies
+        run: yarn --frozen-lockfile
+
+      - name: Publish
+        uses: seek-oss/changesets-snapshot@v0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.SEEK_OSS_CI_NPM_TOKEN }}
```
michaeltaranto, line 31: Happy to use this token, but noticed that we are using a different NPM_TOKEN on the release step. Both should be available, but we should tidy this up.

askoufis: Good catch. I think the preference is to use `SEEK_OSS_CI_NPM_TOKEN`, we swapped to it [in sku not too long ago](https://github.com/seek-oss/sku/pull/700).

michaeltaranto (approved):

Issue #261, opened 2022-06-22
React 18 support


Moranilt:
Playroom does not support React 18 dependencies. It would be greate to add this support.

michaeltaranto:
Yep, started on this a few weeks ago but had to park for a bit. Was playing around with the approach, trying to avoid config if possible.

I'll loop back.

mogularpb:
any news on this? React is no longer required to be imported into components, and in fact recent typescript linting changes mean that you will get an error when you are importing react. However this then breaks playroom.

michaeltaranto:
Picked this up again between tasks and thought i'd share an update:
- Added support for React 18, with backwards compatibility and zero config (drawing inspiration from Storybooks implementation)
- The editor package ([react-codemirror2]) seems to be [no longer updated] and uses legacy React lifecycle APIs which are no longer supported.
- Started working on migration to [Codemirror 6], which is non-trivial requiring all features to be rewired to suit their new extension architecture.

Really want to get this work through, but our effort is definitely focused elsewhere in our ecosystem due to company priorities. I'll try to keep this issue updated.

[react-codemirror2]: https://github.com/scniro/react-codemirror2
[no longer updated]: https://github.com/scniro/react-codemirror2/issues/293
[Codemirror 6]: https://codemirror.net/

michaeltaranto:
Actually, pivoted on this one. Given React still lets you use legacy lifecycle APIs, we have taken in a copy of the editor as a temporary workaround, while we look at upgrading to CodeMirror 6 separately.

This allows us to support React 18 (with backwards compatibility and no config/api change).

https://github.com/seek-oss/playroom/pull/265

PR #265, opened 2022-12-08
fix: Add support for React 18

michaeltaranto:
Upgrade to support using React 18, while also maintaining backwards compatibility and handling React 17.

### React Render API
The approach was heavily inspired by [the approach used by Storybook], while avoiding the need to add config to control the version. Playroom decides which version of the render API to use based on the version of `react-dom` it resolves.

### Code editor
The [react-codemirror2] editor is unmaintained and not compatible with React 18 (using legacy APIs and not throwing the peer dependency warning). As it is just a thin react wrapper with no additional dependencies, and given the legacy lifecycle hooks still work, we opted to embed a copy while we investigate a Codemirror 6 migration separately. 

As part of bringing it in there were some minor modifications:
  - Remove unused Controlled version of the editor
  - Improved type annotations
  
### Extras
Use `@ts-expect-error` declaration instead of `@ts-ignore` and fix handful of issues where types were available.

[the approach used by Storybook]: https://github.com/storybookjs/storybook/pull/17215/files
[react-codemirror2]:  https://github.com/scniro/react-codemirror2

---

Closes https://github.com/seek-oss/playroom/issues/261


askoufis (approved):
image

PR #269, opened 2022-12-12
Update dependencies

michaeltaranto:
Updates all dependencies and remove some extraneous semantic-release and commitizen configuration.


src/Playroom/CodeEditor/CodeEditor.tsx
```diff
@@ -71,7 +71,7 @@ const validateCode = (editorInstance: Editor, code: string) => {
       marker.setAttribute(
         'title',
         // Remove our wrapping Fragment from error message
-        (err.message || '')
+        errorMessage
           .replace(/\<React\.Fragment\>/, '')
           .replace(/\<\/React\.Fragment\>$/, '')
```
askoufis, line 76: I think something might've broken here, seeing `</React.Fragment>` in error messages.
image

michaeltaranto: It appears to be in master too. I'll look into it

.github/workflows/validate.yml
```diff
@@ -21,6 +21,28 @@ jobs:
...
+          path: ${{ steps.pnpm-cache.outputs.STORE_PATH }}
+          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-store-
+
+      - name: Cache Cypress binary
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/Cypress
+          key: cypress-${{ runner.os }}-cypress-${{ hashFiles('**/package.json') }}
```
mrm007, line 42: Should `pnpm-lock.yaml` be used instead of `package.json`? `pnpm-lock.yaml` can change without affecting `package.json`.

mrm007 (approved):

PR #271, opened 2022-12-15
ci: Run changesets release as seek-oss-ci user

michaeltaranto:
Run the changesets release step as the `seek-oss-ci` user.


askoufis (approved):
🤞

PR #273, opened 2023-01-05
Unpin and update `@vanilla-extract/webpack-plugin`

askoufis:
I think `@vanilla-extract/webpack-plugin` accidentally became pinned in #269. In addition, `@vanilla-extract/babel-plugin` was removed in #269, but this requires at least [`@vanilla-extract/webpack-plugin@2.2.0`](https://github.com/vanilla-extract-css/vanilla-extract/releases/tag/%40vanilla-extract%2Fwebpack-plugin%402.2.0), so I've updated the version to `^2.2.0`.

### DIFF
--- a/.changeset/thin-shirts-develop.md
+++ b/.changeset/thin-shirts-develop.md
@@ -1,0 +1,5 @@
+---
+'playroom': patch
+---
+
+Unpin and update `@vanilla-extract/webpack-plugin`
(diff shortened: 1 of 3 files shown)


benjervis (approved):

askoufis:
Seems like there was a good reason to pin the dependency.

Looks like [similar issues have been reported](https://github.com/vanilla-extract-css/vanilla-extract/issues?q=NonErrorEmittedError) in the VE repo.

This is what occurs when starting the basic fixture:
image

PR #277, opened 2023-03-05
Support multi cursors

benjervis:
To add proper multi cursor support, I've upgraded the swap line up/down commands to handle multiple ranges, and have introduced new keybindings to add the extra cursors.

The codemirror file was getting a bit unwieldy with all those key event handlers, so I've split them out into some separate files for organisation.


michaeltaranto:
This is looking sooooo good! I'd like to see some high level integration tests though, like set the code, select next occurrence 4 times, type something, and assert the code in the editor.

Doing this for the various features added would be hugely useful for maintaining all the bindings across code editor implementations.

.changeset/khaki-horses-lay.md
```diff
@@ -0,0 +1,18 @@
...
+---
+
+Adds multi-cursor support.
+
+The keyboard shortcuts added in the previous version (swap/duplicate line up/down) now support multiple cursors being on screen.
+"Select next occurrence" and "add cursor up/down" have also been implemented.
+
+| Keybinding     | Action                  |
+| -------------- | ----------------------- |
+| Alt-Up         | Swap line up            |
```
mrm007, line 12: What do you think about making these shortcuts more semantic / nicer to read?
```suggestion
| <kbd>Alt</kbd>-<kbd>Up</kbd>         | Swap line up            |
```

keyboard

src/Playroom/CodeEditor/CodeEditor.tsx
```diff
@@ -436,6 +267,9 @@ export const CodeEditor = ({ code, onChange, previewCode, hints }: Props) => {
           'Alt-Down': swapLineDown,
           'Shift-Alt-Up': duplicateLine('up'),
           'Shift-Alt-Down': duplicateLine('down'),
+          'Cmd-Alt-Up': addCursorToPrevLine,
+          'Cmd-Alt-Down': addCursorToNextLine,
+          'Cmd-D': selectNextOccurrence,
```
mrm007, line 272: It would be good to have some test coverage for these functions (and the ones above). It's a lot of complex code. What's your confidence that these won't break in a minor CodeMirror version bump? We can't rely just on the @askoufis test suite.

.changeset/khaki-horses-lay.md
benjervis: Oh what I've never even heard of kbd

cypress/e2e/snippets.cy.js
```diff
@@ -67,12 +68,12 @@ describe('Snippets', () => {
 
   it('driven with keyboard', () => {
     // Open and format for insertion point
-    typeCode(`${navigator.platform.match('Mac') ? '{cmd}' : '{ctrl}'}k`);
+    typeCode(`${isMac() ? '{cmd}' : '{ctrl}'}k`);
```
mrm007, line 71: I thought a bit more about this and I think these should be split into two test suites. The shortcuts are different for each platform so it makes sense to test them on each platform.

Also, [some people believe](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-conditional-in-test.md) (myself included) that there should be no conditional logic in tests.

benjervis: Hmm, I think this is a slightly different scenario than just conditional logic. Agree that we should run both versions, but I wouldn't want the tests getting out of sync because they're separate files.

I might just re-run the cypress step in a macos action, so that we know that both versions are working.

.github/workflows/validate.yml
```diff
@@ -55,3 +55,49 @@ jobs:
         run: |
           pnpm cypress:verify
           pnpm cypress
+  test-mac:
+    name: Test on Mac
+    runs-on: macos-latest
```
mrm007, line 60: This can be collapsed with the above [using a matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs).

```suggestion
    runs-on: [ubuntu-latest, macos-latest]
```

benjervis: Requires a slightly more complicated syntax, but that seems to be working

benjervis: Actually no it isn't

mrm007: I think the failure has something to do with the caches. I cleaned all caches and re-ran all jobs.

mrm007: Looks like it's running both Cypresses now

cypress/e2e/snippets.cy.js
benjervis: Feels like this just moves the ternary up a few levels, and will add a bunch of skipped test alerts to the logs? Like we're still doing conditional logic.

cypress/e2e/keymaps.cy.js
```diff
@@ -0,0 +1,187 @@
...
+    loadPlayroom();
+
+    // The last closing div is automatically inserted by autotag
+    typeCode(
+      dedent`
+      <div>First line</div>
+      <div>Second line</div>
+      <div>Third line
+    `
+    );
```
mrm007, line 30:
```suggestion
    typeCode(dedent`
      <div>First line</div>
      <div>Second line</div>
      <div>Third line
    `);
```

mrm007 (approved):
Nice work with the Cypress tests. So much better now 👏

cypress/support/utils.js
```diff
@@ -85,6 +87,20 @@ export const assertFirstFrameContains = (text) => {
...
+export const selectNextCharacters = (numCharacters) => {
+  typeCode('{shift+rightArrow}'.repeat(numCharacters));
+};
+
+export const selectNextWords = (numWords) => {
+  const modifier = isMac() ? 'alt' : 'ctrl';
+  typeCode(`{shift+${modifier}+rightArrow}`.repeat(numWords));
+};
+
+export const selectLines = (numLines, direction = 'down') => {
```
mrm007, line 99:
```suggestion
/**
 * @typedef {import('../../src/Playroom/CodeEditor/keymaps/types').Direction} Direction
 */
/**
 * @param {number}    numLines
 * @param {Direction} direction
 */
export const selectLines = (numLines, direction = 'down') => {
```

.github/workflows/validate.yml
```diff
@@ -35,13 +38,18 @@ jobs:
...
           key: cypress-${{ runner.os }}-cypress-${{ hashFiles('**/pnpm-lock.yaml') }}
           restore-keys: |
             cypress-${{ runner.os }}-cypress-
 
+      - name: Install Cypress
+        if: steps.cypress-cache.outputs.cache-hit != 'true'
+        run: pnpm exec cypress install
+
       - name: Install Dependencies
         run: pnpm i
```
mrm007, line 54:
```suggestion
      - name: Install Dependencies
        run: |
          pnpm install
          pnpm exec cypress install
```

[Seems to be working](https://github.com/seek-oss/playroom/actions/runs/4350117521)

PR #285, opened 2023-06-10
[DO NOT MERGE] Add vite support

TheMightyPenguin:
# Description

Adds vite support. The motivation here:

- To make Playroom easier to adopt. People can pick this up and use it in a TS + React app without any config when using vite as the bundler, they may just need to add a plugin for their CSS solution and that's it.
- Many newer apps build with NextJS/Vite may not have webpack explicitly, so if someone wants to install Playroom now they won't have to deal with webpack. (e.g. the NextJS app may use webpack it behind the curtain but the user maybe does not want to deal with a webpack config).

**This is missing proper build support for vite (it does build but has some errors when running the generated files), and some additional testeing. I'll be working on this over the weekend.**

## Changes mode.

This is a chunky PR (sorry about that), a tl;dr of the changes:

- Refactors the code for the React playroom app to TypeScript. This was done as these files needed to be renamed to .jsx to be supported in Vite, so I thought it would be a good idea to convert these to TypeScript as I was changing these. This also adds a few improvements, like typings for the user defined aliased modules that are used in the codebase.
- Adds a `bundler` option to playroom config. This option is required, and should be either `'vite'` or `'webpack'`.
- Adds the `viteConfig` option to playroom config. This is a function that returns the user Vite configuration.
- Adds a default vite config, following the rules we have in the default webpack config (react support, and this supports TS out of the box 🎉 )


.changeset/fluffy-spies-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'playroom': minor
```
TheMightyPenguin, line 2: tbh not sure if this should be major instead as the new option is required 😅

.babelrc
```diff
@@ -1,6 +1,11 @@
 {
   "presets": [
-    "@babel/preset-typescript",
+    [
+      "@babel/preset-typescript",
+      {
+        "shippedProposals": true
+      }
+    ],
```
TheMightyPenguin, line 8: Comment for me, might need to remove this, I think I broke something in webpack and added this trying to fix it

TheMightyPenguin:
@michaeltaranto this is ready for review! ran cypress tests locally and they are passing

TheMightyPenguin:
@michaeltaranto @askoufis any updates here? 👀  I just rebased the PR as it had some conflicts

TheMightyPenguin:
@michaeltaranto @askoufis following up here!

src/utils/url.ts
```diff
@@ -0,0 +1,96 @@
+import lzString from 'lz-string';
+
+export interface Snippet {
+  group: string;
+  name: string;
+  code: string;
+}
+
+export type Snippets = Snippet[];
```
askoufis, line 9: Perhaps these `Snippet` types should be inside `types.d.ts`. Doesn't feel right that they're alongside URL-related stuff.

src/render.ts
```diff
@@ -7,10 +7,17 @@ const canUseNewReactRootApi =
   reactDomVersion &&
   (reactDomVersion.startsWith('18') || reactDomVersion.startsWith('0.0.0'));
 
-export const renderElement = (node, outlet) => {
+declare let __webpack_public_path__: string | undefined;
```
askoufis, line 10: I don't think `declare` is needed here.
```suggestion
let __webpack_public_path__: string | undefined;
```

src/types.d.ts
```diff
@@ -0,0 +1,48 @@
...
+  snippets?: Snippet[];
+  frameComponent?: string;
+  exampleCode?: string;
+  cwd?: string;
+  storageKey?: string;
+  bundler: 'webpack' | 'vite';
+  viteConfig: () => Promise<any>;
+  webpackConfig?: () => Promise<any>;
+  baseUrl?: string;
+  paramType: 'hash' | 'search';
```
askoufis, line 18: This type can be imported from `url.ts`.

src/types.d.ts
```diff
@@ -0,0 +1,48 @@
...
+  iframeSandbox?: string;
+  // eslint-disable-next-line @typescript-eslint/consistent-type-imports
+  reactDocgenTypescriptConfig?: import('react-docgen-typescript').ParserOptions;
+}
```
askoufis, line 22: Any reason not to declare this as a regular top-level import?

src/render.ts
```diff
@@ -7,10 +7,17 @@ const canUseNewReactRootApi =
...
+
+export const renderElement = async (node: any, outlet: HTMLElement) => {
```
askoufis, line 12: The type of `node` can be improved here:
```suggestion
export const renderElement = async (node: React.ReactNode, outlet: HTMLElement) => {
```

src/types.d.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export interface InternalPlayroomConfig extends PlayroomConfig {
```
askoufis, line 24: Since this is an internal type, should it be in a separate file so that it doesn't get exposed as public API?

src/intersection-observer.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'intersection-observer';
```
askoufis, line 1: Honestly we can probably remove the polyfill for intersection observer since it has pretty broad browser support nowadays. Happy to do this in a followup PR though.

src/Playroom/Frame.tsx
```diff
@@ -1,14 +1,12 @@
-import type React from 'react';
-import type { ReactNode } from 'react';
+import { ComponentType, ReactNode } from 'react';
```
askoufis, line 1: These should still be `type` imports right? I thought we had a lint rule for that...

src/Playroom/Frame.tsx
```diff
@@ -1,14 +1,12 @@
...
 import { useParams } from '../utils/params';
 import CatchErrors from './CatchErrors/CatchErrors';
-// @ts-expect-error
 import RenderCode from './RenderCode/RenderCode';
 
 interface FrameProps {
   themes: Record<string, any>;
-  components: Array<any>;
-  FrameComponent: React.ComponentType<{
+  components: Record<string, ComponentType>;
```
askoufis, line 8: Good catch 👍

.changeset/fluffy-spies-remain.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add support for vite
```
askoufis, line 5: A bit more detail on the breaking change (`bundler` is now required) as well as some info on getting `vite` working would be great here.

cypress/projects/vite/FrameComponent.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
```
askoufis, line 1: Can we convert this, and other "entrypoint" files (e.g. components and snippets) to TypeScript? That way we have an easy way to test out vite + TypeScript support.

lib/makeViteConfig.js
```diff
@@ -0,0 +1,80 @@
...
+      __PLAYROOM_GLOBAL__CONFIG__: JSON.stringify(playroomConfig),
+      __PLAYROOM_GLOBAL__STATIC_TYPES__: JSON.stringify(staticTypes),
+    },
+  };
+
+  const theirConfig = playroomConfig.webpackConfig
+    ? await playroomConfig.viteConfig()
+    : {};
+
+  return vite.mergeConfig(ourConfig, theirConfig);
```
askoufis, line 79: Would this work?
```suggestion
  const theirConfig = await playroomConfig?.viteConfig?.();

  return vite.mergeConfig(ourConfig, theirConfig);
```

askoufis (commented):
Thanks for the amazing effort! 👏

I gave this a crack a while ago, but didn't get too far. IMO the ultimate version of vite support for playroom would be a plugin, rather than a separate CLI + devserver, though that would likely involve a complete re-architect of playroom, so this is a very welcome addition in the meantime.

Just some minor comments from me.

TheMightyPenguin:
@askoufis would you elaborate more on this plugin idea? (I am guessing it means there would be a plugin system and there would be webpack, vite plugins and so on)

Also thanks for the review! gonna get to attending it this weekend!

.changeset/fluffy-spies-remain.md
TheMightyPenguin: agreed good catch, I love a good detailed changeset so will add it

src/Playroom/Frame.tsx
TheMightyPenguin: yes good point, not sure why I removed the `type` import

askoufis:
> @askoufis would you elaborate more on this plugin idea? (I am guessing it means there would be a plugin system and there would be webpack, vite plugins and so on)
> 
> Also thanks for the review! gonna get to attending it this weekend!

Yeah something like that. Basically, rather than playroom being a CLI that spins up a dev server, it would be a plugin that would integrate with bundlers like webpack and vite, using their dev servers to host the playroom site.

src/utils/url.ts
TheMightyPenguin: makes sense, moved

src/render.ts
TheMightyPenguin: removed

src/types.d.ts
TheMightyPenguin: done

src/render.ts
TheMightyPenguin: good catch, changed!

TheMightyPenguin: @askoufis changed but had to left some casting in as we still support older react versions that use ReactDOM.render and that needs a diff type

src/types.d.ts
TheMightyPenguin: good point, sounds like a good idea, moving

src/intersection-observer.d.ts
TheMightyPenguin: yeah will do so in a follow up!

src/Playroom/Frame.tsx
TheMightyPenguin: @askoufis the lint rule is there, I think it was added in other PR that moved some things to TypeScript and I forgot to fix that when I rebased!

cypress/projects/vite/FrameComponent.jsx
TheMightyPenguin: yeah good idea, changing!

lib/makeViteConfig.js
TheMightyPenguin: yeah it should, changing (and testing it)

TheMightyPenguin:
@askoufis attended comments!

src/types.d.ts
TheMightyPenguin: not really, moved!

package.json
```diff
@@ -115,6 +122,7 @@
     "@changesets/cli": "^2.25.2",
     "@octokit/rest": "^19.0.5",
     "@types/jest": "^29.2.4",
+    "@types/webpack-env": "^1.18.4",
```
askoufis, line 125: Just want to confirm, does this package provide types for the webpack HMR API?

src/render.ts
```diff
@@ -0,0 +1,28 @@
...
+let __webpack_public_path__: string | undefined;
+
+export const renderElement = async (node: ReactNode, outlet: HTMLElement) => {
+  if (canUseNewReactRootApi) {
+    // webpack needs to know the public path when doing dynamic imports,
+    // otherwise the HTML chunk for the preview page, will end up importing
+    // react-dom/client from the wront path
+    if (typeof __webpack_public_path__ !== 'undefined') {
+      __webpack_public_path__ = '../../../';
+    }
```
askoufis, line 20: What's the reasoning behind this addition?

askoufis (commented):
Looking good. Just a few more small comments.

src/types.d.ts
```diff
@@ -0,0 +1,31 @@
...
+  exampleCode?: string;
+  cwd?: string;
+  storageKey?: string;
+  bundler: 'webpack' | 'vite';
+  viteConfig?: () => Promise<any>;
+  webpackConfig?: () => Promise<any>;
+  baseUrl?: string;
+  paramType: ParamType;
+  iframeSandbox?: string;
+  // eslint-disable-next-line @typescript-eslint/consistent-type-imports
```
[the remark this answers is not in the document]
TheMightyPenguin: yes, removed

src/render.ts
```diff
@@ -0,0 +1,28 @@
...
+  reactDomVersion &&
+  (reactDomVersion.startsWith('18') || reactDomVersion.startsWith('0.0.0'));
+
+let __webpack_public_path__: string | undefined;
+
+export const renderElement = async (node: ReactNode, outlet: HTMLElement) => {
+  if (canUseNewReactRootApi) {
+    // webpack needs to know the public path when doing dynamic imports,
+    // otherwise the HTML chunk for the preview page, will end up importing
+    // react-dom/client from the wront path
```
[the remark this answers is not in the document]
TheMightyPenguin: good catch!

michaeltaranto (commented):
Thanks for putting in this together. I would like to see this progress, but really feel like this would be easier to support if it were broken into separate PRs.

As I see it we have:
- Migrate internals to TS
- JSX support
- Vite support

Currently the `utils` refactor breaks a public (not documented) [API]:
```ts
import { createUrl } from 'playroom/utils';
```
We use this to build Playroom links on the Braid docs site.

Also would be keen to explore not having the `bundler` config, and using the presence of `webpackConfig` or `viteConfig` as the decision point. This would avoid breaking existing configs too.

[API]: https://github.com/seek-oss/playroom/pull/82

simon-v-swyftx:
Thanks very much @TheMightyPenguin !
I was trying to find some prior art for playroom config with tailwindcss/postcss support, unfortunately the projects I found were out of date or I simply couldn’t get it working with the same/like configuration.
But I pointed to your branch and worked right out of the box! :tada: :heart:

mihkeleidast:
Anything we can do to push this forward?

ccombe:
whatever happened with this PR - webpack is getting very long in the tooth and vite is becoming if not already the standard build tool - what is blocking this change from making it in - other than perhaps a rebase / update to latest versions.

ccombe:
thanks @felixhabib appreciate the quick response 

I see @michaeltaranto suggest attacking this from 2 separate prs with one being
`Migrate internals to TS + JSX support`

how has that part gone (if at all) maybe it's easier to migrate that first - I imagine some of the simpler migration could be done via AI and then reviewed but would need more scrutiny for the more complicated steps.

felixhabib:
@ccombe yes, the TS migration has already been completed separately.

PR #287, opened 2023-07-10
Upgrade to TypeScript 5.0+ and drop support for React 16

mrm007:
I've also upgraded the ESLint config and autofixed a few errors.

The tricky part is ensuring everything can be type checked correctly, as Cypress globals pollute the global namespace. So now we have two separated type checking processes: one for `cypress/` and one for everything else, which works quite nicely (and it's also what [the Cypress docs recommend](https://docs.cypress.io/guides/tooling/typescript-support#Configure-tsconfigjson)).


lib/makeWebpackConfig.js
```diff
@@ -175,16 +164,6 @@ module.exports = async (playroomConfig, options) => {
...
-      // If using a version of React earlier than 18, ignore the
-      // react-dom/client import. This hack can be removed when
-      // support for older versions of React is removed.
-      ...(isLegacyReact
-        ? [
-            new webpack.IgnorePlugin({
-              resourceRegExp: /react-dom\/client$/,
-            }),
-          ]
-        : []),
```
askoufis, line 187: Isn't this still required for react 17 compatibility?

lib/makeWebpackConfig.js
```diff
@@ -19,22 +18,6 @@ module.exports = async (playroomConfig, options) => {
...
-        encoding: 'utf-8',
-      }
-    );
-    const { version } = JSON.parse(pkgContents);
-    isLegacyReact = !(version.startsWith('18') || version.startsWith('0.0.0'));
-  } catch (e) {
-    throw new Error('Unable to read `react-dom` package json');
-  }
-
   const staticTypes = await getStaticTypes(playroomConfig);
```
askoufis, line 21: Isn't this still required for react 17 compatibility?

src/render.js
```diff
@@ -1,19 +1,6 @@
...
 
 export const renderElement = (node, outlet) => {
-  if (canUseNewReactRootApi) {
-    // eslint-disable-next-line import/no-unresolved
-    const { createRoot } = require('react-dom/client');
-    const root = createRoot(outlet);
-    root.render(node);
-  } else {
-    ReactDOM.render(node, outlet);
-  }
```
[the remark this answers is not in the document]
mrm007: It's required. Got a little trigger happy.

askoufis (approved):

PR #288, opened 2023-07-10
Add a TypeScript fixture

mrm007:


cypress/projects/typescript/components/Bar.tsx
```diff
@@ -0,0 +1,13 @@
+import type React from 'react';
+import { parent } from './styles';
+
+type Props = {
+  color: 'red' | 'blue';
```
askoufis, line 5:
```suggestion
  color: 'red' | 'blue' | 'black';
```

cypress/projects/typescript/components/Foo.tsx
```diff
@@ -0,0 +1,13 @@
+import type React from 'react';
+import { parent } from './styles';
+
+type Props = {
+  color: 'red' | 'blue';
```
askoufis, line 5:
```suggestion
  color: 'red' | 'blue' | 'black';
```

cypress/projects/typescript/playroom.config.js
```diff
@@ -0,0 +1,31 @@
...
+          use: [
+            {
+              loader: require.resolve('babel-loader'),
+              options: {
+                presets: [
+                  require.resolve('@babel/preset-env'),
+                  [
+                    require.resolve('@babel/preset-react'),
+                    { runtime: 'automatic' },
+                  ],
```
askoufis, line 22: Do we need `preset-typescript` too?

mrm007: It seems to be working without it. But I added it anyway.

mrm007: It works without it because the main Webpack config handles it https://github.com/seek-oss/playroom/blob/720d542ab73d64573efccea8df77c5a8d7c8be0e/lib/makeWebpackConfig.js#L82-L83

askoufis: Ah

mrm007: It was not that, it was the root `.babelrc` that was making things work. Confirmed it doesn't work without `@babel/preset-react` in the [latest commit](https://github.com/seek-oss/playroom/pull/288/commits/ec0b0d7574a4e8758782f4d9ea1784a0a701012f).

askoufis (approved):

PR #291, opened 2023-07-27
Allow overriding Webpack module rules

mrm007:
See changeset for details.

### DIFF
--- a/.changeset/tidy-cows-complain.md
+++ b/.changeset/tidy-cows-complain.md
@@ -1,0 +1,23 @@
+---
+'playroom': patch
+---
+
+Allow overriding Webpack module rules
+
+Consumers may have complex Webpack configurations that can clash with Playroom's.
+In this case, it's useful to be able to override the module rules that Playroom defines.
+For example, overriding loaders defined for CSS files:
+
+```js
+// playroom.config.js
+module.exports = {
+  webpackConfig: () => ({
+    module: {
+      rules: [
+        // use your own CSS loaders
+        { test: /\.css$/, use: ['style-loader', 'css-loader'] },
+      ],
+    },
+  }),
+};
+```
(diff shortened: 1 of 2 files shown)


askoufis (approved):

pascalduez:
Hello,

this should probably have been released as a major version? 
It breaks our current implementation.

```sh
ERROR in ./components/FrameComponent.js 11:4

Module parse failed: Unexpected token (11:4)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| export default function Root(props) {
|   return (
>     <div>
```

Investigating.

mrm007:
Hi @pascalduez,

Overriding Playroom's Webpack loaders means you have to provide support for processing JSX in your JS/TS files. Playroom merges its default config with yours, overwriting Playroom's `module.rules` with yours. This is an escape hatch for when you need to do something more complex than [what Playroom provides out of the box](https://github.com/seek-oss/playroom/blob/a04486453b7746a1cb95749e0d41db0e152119e6/lib/makeWebpackConfig.js#L80-L144). The caveat is that you have to recreate what Playroom provides out of the box.

If you have a reproducible example of this not working, I'd be happy to take a look.

rbardini:
This change breaks my build too:

```
Module parse failed: Unexpected token (1:0)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
> .storybook-button {
|   font-family: 'Nunito Sans', 'Helvetica Neue', Helvetica, Arial, sans-serif;
|   font-weight: 700;
```

even though I have the following rule:

```js
{
  test: /\.css$/,
  include: path.resolve(__dirname, 'src'),
  use: ['style-loader', 'css-loader'],
}
```

v0.32.0 works with the config above.

PR #299, opened 2023-11-02
Feat: add support for loading mjs config files

nib-bturner:
As a consumer, I want to be able to define my playroom config using ESM compatible JS code. I noticed today when trying this behaviour out that sadly playroom does not currently support this, so I've prepared this PR to add support for this behaviour.

As this PR replaces `require` with `import` for loading config files, it is important that the `bin/cli.js` file is still loaded as CommonJS code. To force this behaviour (when users might specify `"type":"module"` in their `package.json`s) I have renamed it `bin/cli.cjs`.

Unsure if this is best released as a minor or major change. I can modify the changeset command as needed.

### DIFF
--- a/.changeset/lazy-boxes-speak.md
+++ b/.changeset/lazy-boxes-speak.md
@@ -1,0 +1,19 @@
+---
+'playroom': minor
+---
+
+Add support for loading mjs config files
+
+Consumers should now be able to write their configuration files using ES modules. This can be achieved by:
+
+1. **Recommended approach**: by renaming their `playroom.config.js` => `playroom.config.mjs` and specifying the `--config` option when running playroom. For example:
+
+```sh
+  npx playroom --config ./playroom.config.mjs
+```
+
+1. By specifying `"type": "module"` in the package.json, any `.js` files are considered as ES modules. Hence you should be able to load your playroom config without the need to specify the configuration path. ie:
+
+```sh
+  npx playroom
+```
--- a/.changeset/lazy-boxes-speak.md
+++ b/.changeset/lazy-boxes-speak.md
@@ -4,16 +4,4 @@
 
 Add support for loading mjs config files
 
-Consumers should now be able to write their configuration files using ES modules. This can be achieved by:
-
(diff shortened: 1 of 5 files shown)


nib-bturner:
Hi @askoufis,
I noticed you're the most recent reviewer on this project. Is there any chance you can have a look over this feature request? 

Thanks 🙏🏼

bin/cli.cjs
askoufis, line 1: `package.json` will need to be updated to reflect the file rename.
https://github.com/seek-oss/playroom/blob/9de7d5ec663bdbdae3dc8084a68f442a86dd0438/package.json#L8

askoufis (commented):
A minor release is fine for this one. We're staying 0-versioned for now.

Would also be great to add a little note about ESM config in the README too. Doesn't have to be a full example, but just worth mentioning that an explicit `--config` flag might be required in some cases.

bin/cli.cjs
```diff
@@ -62,7 +62,7 @@ const showUsage = () => {
     process.exit(1);
   }
 
-  const config = require(configPath);
+  const { default: config } = await import(configPath);
```
askoufis, line 65: Actually, we can pass an array of files to `findUp` on line 56 and then consumers don't need to worry about passing an explicit `--config` flag, unless they rename their config.

```
await findUp(['playroom.config.js', 'playroom.config.mjs', 'playroom.config.cjs'], { cwd })
```

nib-bturner: I have not used `findUp` before - but this makes way more sense. I've also adjusted the README to reflect these changes.

.changeset/lazy-boxes-speak.md
askoufis, line 1: The changeset can be tweaked a bit now that `findUp` is handling the different file extensions.

askoufis:
Looks like [the eslint config](https://github.com/seek-oss/playroom/blob/9de7d5ec663bdbdae3dc8084a68f442a86dd0438/.eslintrc#L9) needs to be tweaked, but then this should be good to go.

askoufis (approved):
Tested this change in a separate project and confirmed that it functions as expected. Thanks for the contribution!

PR #301, opened 2023-11-12
Replace `@babel/standalone` with `sucrase` for JSX compilation

askoufis:
[`sucrase`](https://github.com/alangpierce/sucrase) is a much faster alternative to babel. It gets its speed by only supporting modern browsers and versions of node, and sacrificing extensibility. It supports JSX out of the box, and runs in the browser, so this was basically a drop-in replacement for `@babel/standalone`.

## Performance

For small playrooms, the difference in execution time of `compileJsx` is pretty negligible. Both `@babel/standalone` and `sucrase` often compile in <10ms. However, in larger playrooms (tested a [Braid](https://github.com/seek-oss/braid-design-system) playroom with ~2300 lines of code), `@babel/standalone` often takes between 100-200ms to compile, whereas `sucrase` still compiles in ~10ms. This results in a noticeable difference to input latency and general responsiveness.

It seems like snippet performance (opening the snippet menu and previewing snippets) has improved a bit too.

Playroom URLs in `.txt` files because they're too long for github and URL shorteners. Try typing quickly on line 328.
- [braid master large playroom URL.txt](https://github.com/seek-oss/playroom/files/13327877/braid.master.large.playroom.URL.txt)
- [sucrase snapshot large braid playroom URL.txt](https://github.com/seek-oss/playroom/files/13327876/sucrase.snapshot.large.braid.playroom.URL.txt)

**NOTE**: These profile were taken from the dev server playroom because it was easier than setting up prod source maps. The compilation time difference is probably a bit smaller in prod build, but the difference in responsiveness is still clearly noticeable in the playroom links above.

[Chrome performance profiles](https://github.com/seek-oss/playroom/files/13327170/profiles.zip)

## Bundle size

`@babel/standalone` is a chunky dependency. `sucrase` is comparatively much lighter.
- [Current master Braid playroom: ~1.16MB gzipped](https://braid-design-system--561d50f28f85a3828cfa3b4356226003ff179eec.surge.sh/playroom/#?code=N4IgLgFgpgtlDOIBcBtE8pQNYCkD2ARogLoA0IA7gJYAmkiqAzAEwAMxAvkA)
- [Sucrase snapshot playroom: ~611kB gzipped](https://braid-design-system--e4947591b5c3bfa14c375c8e740237b4a8900953.surge.sh/playroom/#?code=N4IgLgFgpgtlDOIBcBtE8pQNYCkD2ARogLoA0IA7gJYAmkiqAzAEwAMxAvkA)

## Future

There's some more stuff we can improve on in the short term, such as using custom JSX pragmas to reduce the amount of code being sent to each iframe (will do a followup PR). We could also try using [`esbuild-wasm`](https://esbuild.github.io/api/#browser), but its API is async only, so it would involve a bit of refactoring. It's also somewhat slower than `sucrase`, according to the benchmarks in `sucrase`'s README.

Since we don't specify targets in `babel-loader`, we're [compiling down for the oldest browser possible](https://babeljs.io/docs/options#no-targets). AFAIK Playroom has never explicitly advertised its supported browser versions. If we moved to `sucrase` for the webpack loader we could potentially get some build time speedups, at the cost of only supporting relatively modern browser versions.


src/index.d.ts
```diff
@@ -15,6 +15,7 @@ interface PlayroomConfig {
   iframeSandbox?: string;
   // eslint-disable-next-line @typescript-eslint/consistent-type-imports
   reactDocgenTypescriptConfig?: import('react-docgen-typescript').ParserOptions;
+  production?: boolean;
```
mrm007, line 18: I don't think this is a thing.
```suggestion
```

askoufis: It's specified in the `start` and `build` scripts.

https://github.com/seek-oss/playroom/blob/26828a5a64f7774ed63b60222ba1f6eabe7dbaab/lib/start.js#L10
https://github.com/seek-oss/playroom/blob/26828a5a64f7774ed63b60222ba1f6eabe7dbaab/lib/build.js#L6

mrm007: The first argument is the playroom config. The second argument is not documented/typed anywhere.

mrm007: I don't see where these other options are used.

https://github.com/seek-oss/playroom/blob/a7e478dcc0d5e38f29a04bbf126188c4c2647bfb/lib/start.js#L11-L16

https://github.com/seek-oss/playroom/blob/a04486453b7746a1cb95749e0d41db0e152119e6/lib/makeWebpackConfig.js#L18

askoufis: Ah, good point.

mrm007 (approved):

michaeltaranto (approved):
Looks good to me on testing of some of the more complex Playrooms.

PR #302, opened 2023-11-16
:sparkles: add new config variable "defaultWidths"

JulianNymark:
# What and Why?

Allow default widths to be set to something other than "show all the widths available", since we support a lot of breakpoints, but for the average developer using our design system it's a bit overwhelming to see all of them + horizontal scroll when they are _usually_ just interested in one or two breakpoints (for us it's usually a mobile width + generic desktop widths).

image

# How?

introduces a configuration variable `defaultWidths` that becomes the viewports rendered when you have no width check mark selected in the sidebar (as shown in the screenshot above).

```javascript
module.exports = {
  // ...
  widths: [320, 480, 768, 1024, 1280, 1440],
  defaultWidths: [320, 1280]
};

```

### DIFF
--- a/src/Playroom/Playroom.tsx
+++ b/src/Playroom/Playroom.tsx
@@ -50,10 +50,11 @@ export interface PlayroomProps {
   components: Record<string, ComponentType>;
   themes: string[];
   widths: number[];
+  defaultWidths: number[];
   snippets: Snippets;
 }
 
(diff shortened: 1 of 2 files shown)


michaeltaranto:
Addressed by https://github.com/seek-oss/playroom/pull/305

Issue #176, opened 2020-07-27
Initial brands and screen sizes config


lauriejones:
### Feature request

Similar to how you can provide the initial example code via config, I think it would be great to be able to specify the initially enabled brands and screen sizes.

We have one theme that is the primary brand, and then 8 others that are very much secondary. I'd love to be able to set just the primary brand to checked initially to give it focus for the first impression.

askoufis:
Implemented in https://github.com/seek-oss/playroom/pull/305.

PR #305, opened 2023-11-27
Allow specifying defaults for visible themes and widths

nib-bturner:
This PR adds support for specifying default screen sizes and themes to be rendered when a state is not found in the query string or from being already specified in the StoreProvider react context.

Addresses: https://github.com/seek-oss/playroom/issues/176


nib-bturner:
Morning @askoufis,

Seems like there is quite some demand for the ability to specify a default theme and screen widths (eg https://github.com/seek-oss/playroom/pull/302). I am hoping you might be able to have a look over this PR, so we can start to use some of these features. Thanks!

nib-bturner:
@michaeltaranto any chance you can have a look over this one? We support about 20 different themes across ~6 different screen sizes. Initially loading all of these is very overwhelming and we would love to specifying defaults to prevent this cognitive overload.

src/index.js
```diff
@@ -38,7 +38,12 @@ polyfillIntersectionObserver().then(() => {
     );
 
     renderElement(
-      <StoreProvider themes={themeNames} widths={widths}>
+      <StoreProvider
+        themes={themeNames}
+        widths={widths}
+        defaultVisibleThemes={playroomConfig.defaultVisibleThemes}
+        defaultVisibleWidths={playroomConfig.defaultVisibleWidths}
```
michaeltaranto, line 45: Is there a reason to pass the config through to the provider from here? I ask because we already [import the config] in the store itself to setup some initial state.

[import the config]: https://github.com/seek-oss/playroom/blob/e2bd10e0f7f0d0509e352ebb4d0ad354c73f44f3/src/StoreContext/StoreContext.tsx#L19

README.md
```diff
@@ -80,6 +80,12 @@ module.exports = {
     // Custom webpack config goes here...
   }),
   iframeSandbox: 'allow-scripts',
+  defaultVisibleWidths: [
+    // Custom subset of your specified widths to display on first load
+  ],
+  defaultVisibleThemes: [
+    // Custom subset of your specified themes to display on first load
+  ],
```
michaeltaranto, line 88: Nit pick, but some bikeshedding:
```suggestion
  defaultVisibleWidths: [
    // subset of widths to display on first load
  ],
  defaultVisibleThemes: [
    // subset of themes to display on first load
  ],
```

.changeset/itchy-taxis-rhyme.md
```diff
@@ -0,0 +1,5 @@
+---
+'playroom': minor
+---
+
+Add support for specifying default subsets of themes and screen widths
```
michaeltaranto, line 5: Maybe a nice touch to add the docs snippet here:

````suggestion
Add support for specifying default subsets of themes and screen widths via the config.

#### Example usage:
```js
// playroom.config.js
module.exports = {
  ...,
  defaultVisibleWidths: [
    // subset of widths to display on first load
  ],
  defaultVisibleThemes: [
    // subset of themes to display on first load
  ],
}
```
````

michaeltaranto (commented):
Just a few tidy ups to called out. Thanks for this

src/index.js
nib-bturner: Ah I missed that playroom config was imported in the StoreContext. I've adjusted the logic to consume it there rather than passing it in here.

michaeltaranto (approved):
Looks pretty clean mate. Thank you!

tob:
This feature is great, thanks for the PR, just a small note, it could be worth to mention in the readme that to the defaultVisibleThemes you don't pass the theme object but just the theme name.

PR #308, opened 2023-12-12
Make the error marker hoverable

mrm007:
Currently there's no way to tell that there's a tooltip on the error marker, other than waiting

### DIFF
--- a/.changeset/thick-mails-poke.md
+++ b/.changeset/thick-mails-poke.md
@@ -2,4 +2,4 @@
 'playroom': patch
 ---
 
-Make the error marker hoverable
+Improve affordance of error marker detail
--- a/.changeset/thick-mails-poke.md
+++ b/.changeset/thick-mails-poke.md
@@ -1,0 +1,5 @@
+---
+'playroom': patch
+---
+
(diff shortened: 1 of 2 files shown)


src/Playroom/CodeEditor/CodeEditor.css.ts
```diff
@@ -23,6 +23,9 @@ export const errorMarker = style([
     opacity: 0,
   }),
   {
+    ':hover': {
+      cursor: 'pointer',
+    },
```
askoufis, line 28: Should `cursor: 'pointer'` be used when there isn't really a mouse action to be performed? Maybe a hover style would be more appropriate?

.changeset/thick-mails-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+'playroom': patch
+---
+
+Make the error marker hoverable
```
michaeltaranto, line 5: It is currently "hover-able", so maybe something like:

```suggestion
Improve affordance of error marker detail
```

Thoughts?

askoufis (approved):

Issue #311, opened 2024-01-02
Unable to build playroom without a favicon


jordanoverbye:
I'm unable to run `playroom build` after upgrading to `0.34.1`. I'm getting the following error:

```
Error: HtmlWebpackPlugin: could not load file /path/to/repo/images/favicon.png
- index.js:494      [repo]/[html-webpack-plugin]/index.js:494:35     error Command failed with exit code 1.
```

This seems to have been caused by PR #309

If I create a `images/favicon.png` file everything works correctly, but this seems unexpected for a patch release.

mrm007:
@jordanoverbye does [this snapshot](https://github.com/seek-oss/playroom/actions/runs/7635268557) work for you?

PR #312, opened 2024-01-02
Add new `faviconPath` option

jordanoverbye:
This PR fixes issue #311 (which appears to be breaking builds in version `0.34.1`) by adding a new `faviconPath` option.

I've also tried to simplify the favicon implementation by replacing the `favicon.png` and `favicon-inverted.png` files with a single [SVG favicon](https://css-tricks.com/svg-favicons-and-all-the-fun-things-we-can-do-with-them/).


lib/makeWebpackConfig.js
```diff
@@ -37,6 +37,11 @@ module.exports = async (playroomConfig, options) => {
 
   const staticTypes = await getStaticTypes(playroomConfig);
 
+  // Use the favicon from the config if specified, otherwise don't use anything
+  const favicon = playroomConfig.faviconPath
+    ? relativeResolve(playroomConfig.faviconPath)
+    : undefined;
```
jordanoverbye, line 43: I've defaulted to `undefined` here because I believe the /images folder is not published to npm, and I'm not sure if you want to do that. However, I'm happy to change this if needed.

askoufis:
SVG icons aren't supported at all in Safari, and some mobile browsers. I would suggest that it's not worth worse browser support for a slight simplification. This issue can be fixed without also adding a feature too.

PR #313, opened 2024-01-06
Migrate playroom internals to TypeScript

TheMightyPenguin:
Migrates internals to TypeScript. This PR is the first in a series of changes aiming to support Vite. This is the original PR https://github.com/seek-oss/playroom/pull/285, but following Michael suggestion to split up the work.


TheMightyPenguin:
@michaeltaranto @askoufis breaking down the Vite PR into smaller ones, this migrates files in the `src/` folder to TypeScript, still somewhat chunky but should be easier to review as it should maintain the same current Playroom behavior.

package.json
```diff
@@ -2,8 +2,8 @@
   "name": "playroom",
   "version": "0.34.1",
   "description": "Design with code, powered by your own component library",
-  "main": "utils/index.js",
-  "types": "utils/index.d.ts",
+  "main": "utils/index.ts",
+  "types": "utils/index.ts",
```
askoufis, line 6: Do we want to ship typescript as source? The alternative would be a build step. We could just run `tsc` for now, and graduate to something that offers proper CJS/ESM support later.

src/Playroom/Frame.tsx
```diff
@@ -1,19 +1,18 @@
...
 import { useParams } from '../utils/params';
 import CatchErrors from './CatchErrors/CatchErrors';
-// @ts-expect-error
 import RenderCode from './RenderCode/RenderCode';
 
 interface FrameProps {
   themes: Record<string, any>;
-  components: Array<any>;
-  FrameComponent: React.ComponentType<{
+  components: Record<string, ComponentType>;
```
askoufis, line 8: Should this be `ComponentType<any>` so it will accept any component? Leaving out the generic argument results in it defaulting to `{}`, so only components that don't take any props would be accepted.

src/Playroom/RenderCode/RenderCode.tsx
```diff
@@ -9,7 +9,13 @@ import {
   ReactFragmentPragma,
 } from '../../utils/compileJsx';
 
-export default function RenderCode({ code, scope }) {
+export default function RenderCode({
+  code,
+  scope,
+}: {
+  code: string | undefined;
+  scope: Record<string, ComponentType>;
```
askoufis, line 17: I don't think `ComponentType` is correct here. You can inject whatever you want with `scope`, a constant, function, component, etc. Maybe `Record<string, any>` would be more appropriate.

askoufis (commented):
Thanks for splitting out the PR. Just a few comments.

package.json
askoufis: Reflecting on this, maybe for now we could just keep the separate `.js` and `.d.ts` files for simplicity, and think about a potential build process later on.

michaeltaranto:
Closing this PR, as we broke the work down into https://github.com/seek-oss/playroom/pull/422, https://github.com/seek-oss/playroom/pull/421, https://github.com/seek-oss/playroom/pull/419 and https://github.com/seek-oss/playroom/pull/418.

Thanks so much for all the effort in this space, apologies we have not been able to focus on this earlier.

❤️

TheMightyPenguin:
@michaeltaranto thanks! I mean I also haven't follow up in like a year so appreciate seeing this making its way into the repo in some way 😄

PR #319, opened 2024-02-01
Allow configuring of tab titles

felixhabib:
It is hard to distinguish between Playroom tabs when there is more than one in the same window.
image

This change:
* Adds a "Title" section in the settings pane which allows you to customise the tab title.
* Saves Titles in the URL.
* Updates spacing scale logic.
* Removes dividers in all menus for consistent visual treatment.

# For example
image


src/Playroom/SettingsPanel/SettingsPanel.tsx
```diff
@@ -73,101 +74,154 @@ const KeyboardShortcut = ({
...
+              onKeyDown={(e) => {
+                if (e.key === 'Enter') {
+                  dispatch({
+                    type: 'toggleToolbar',
+                    payload: { panel: 'settings' },
+                  });
+                }
+              }}
+            />
+          </fieldset>
```
michaeltaranto, line 133: Given title is just one field, this should just be an input and label rather than a fieldset. We can run through this together separately.

src/Playroom/SettingsPanel/SettingsPanel.tsx
```diff
@@ -73,101 +74,154 @@ const KeyboardShortcut = ({
...
-                  </span>
-                </label>
+    <>
+      {localTitle === undefined ? null : (
+        <Helmet>
+          <title>{displayedTitle}</title>
+        </Helmet>
+      )}
+      <ToolbarPanel data-testid="frame-panel">
+        <Stack space="small" dividers>
```
michaeltaranto, line 107: The divider looks a bit close, this should possibly be large like [here] but we can play with this when we update the field (as below).

[here]: https://github.com/seek-oss/playroom/pull/319/files#diff-4021ca4ad6a2db46421d46681ae15da803ba3543b1cee9ba7100027136eca0c8R135

src/Playroom/SettingsPanel/SettingsPanel.tsx
```diff
@@ -73,101 +74,144 @@ const KeyboardShortcut = ({
...
-                  </span>
-                </label>
-              </div>
+    <>
+      {localTitle === undefined ? null : (
+        <Helmet>
+          <title>{displayedTitle}</title>
+        </Helmet>
+      )}
+      <ToolbarPanel data-testid="frame-panel">
```
michaeltaranto, line 106: Minor copy and paste issue (not from your PR)
```suggestion
      <ToolbarPanel data-testid="settings-panel">
```

felixhabib: Ahh, nice catch!

utils/index.js
```diff
@@ -1,11 +1,12 @@
 const lzString = require('lz-string');
 
-const compressParams = ({ code, themes, widths, theme }) => {
+const compressParams = ({ code, themes, widths, theme, title }) => {
   const data = JSON.stringify({
     ...(code ? { code } : {}),
     ...(themes ? { themes } : {}),
     ...(widths ? { widths } : {}),
     ...(theme ? { theme } : {}),
+    ...(title ? { title } : {}),
```
michaeltaranto, line 9: Noticed we are not passing the `title` via the `createUrl` function below. This could be a cool addition allow URLs to be generated with pre-defined titles.

felixhabib: Good idea, should be added now!

src/Playroom/SettingsPanel/SettingsPanel.css.ts
```diff
@@ -98,3 +98,25 @@ export const label = style([
...
+    boxSizing: 'border-box',
+    borderRadius: 'medium',
+  }),
+  {
+    color: colorPaletteVars.foreground.neutral,
+    height: vars.touchableSize,
+    background: colorPaletteVars.background.surface,
+    ':focus': {
+      outline: 'none',
+    },
```
askoufis, line 116: Any particular reason to disable the focus outline?

felixhabib: I think this might be a leftover property from before we put it in a box and maybe should be removed now. @benjervis should be able to provide more context.

felixhabib: Yes this was the case, I'll remove it. Nice catch!

benjervis: I think this was covering for the implementation of my original mediocre design, but with the way it looks now I think the focus outline is fine and we can put it back

michaeltaranto (approved):

PR #321, opened 2024-02-13
Add comment command

felixhabib:


askoufis (commented):
Did a high-level pass, will do a more detailed on when it's not in draft 😂

package.json
```diff
@@ -139,6 +139,7 @@
   },
   "packageManager": "pnpm@8.15.3",
   "volta": {
-    "node": "18.19.1"
+    "node": "18.19.1",
+    "pnpm": "7.18.1"
```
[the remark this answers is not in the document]
felixhabib: Ahh right. I'll remove the change.

cypress/e2e/keymaps.cy.js
```diff
@@ -314,4 +349,1450 @@ describe('Keymaps', () => {
...
+        prop2="This is the second prop"
+        prop3="This is the third prop"
+      >
+        First line
+      </div>
+      <div>Second line</div>
+      <div>Third line</div>`;
+
+    const modifierKey = isMac() ? 'cmd' : 'ctrl';
+    const typeComment = () => typeCode(`{${modifierKey}+/}`);
```
askoufis, line 371: I initially interpreted this as "type a comment", but really it's doing "execute comment hotkey" or something of that nature.
```suggestion
    const executeCommentHotkey = () => typeCode(`{${modifierKey}+/}`);
```

cypress/e2e/keymaps.cy.js
```diff
@@ -314,4 +349,1450 @@ describe('Keymaps', () => {
...
+        });
+
+        it('line', () => {
+          loadPlayroom(lineStarter);
+
+          moveBy(0, 1);
+          moveToEndOfLine();
+
+          // Todo - (1/2) Solve issue where Ubuntu does not select to the start of line
+          // Todo - (2/2) with one trigger of the keybinding
```
askoufis, line 540: Has this been addressed? Is this an ubuntu issue, or a broader linux issue?

src/Playroom/CodeEditor/keymaps/comment.ts
```diff
@@ -0,0 +1,325 @@
...
+        const uncommentType: CommentType = existingContent
+          .trimStart()
+          .startsWith(BLOCK_COMMENT_START)
+          ? 'block'
+          : 'line';
+
+        const existingContentWithoutComment = existingContent.replace(
+          uncommentType === 'block'
+            ? /\{\/\*\s?|\s?\*\/\}/g
+            : /^(\s*)\/\/\s?/gm,
```
askoufis, line 294: Not super concerned about perf TBH given the testing we've done, but it's still worth pulling any regexes out of this function into top-level consts, both so they don't get created each time the operation is run, and also to give them names so it's obvious what they're for.

cypress/e2e/keymaps.cy.js
felixhabib: Yep good idea

cypress/e2e/keymaps.cy.js
felixhabib: This hasn't been addressed. I'm not 100% sure who the culprit is but by default on the Ubuntu tests, the backwards selection goes up to the opening "//" (not including), whereas macOS selection goes to the very start of the line.

src/Playroom/CodeEditor/keymaps/comment.ts
```diff
@@ -0,0 +1,329 @@
...
+        const updatedContent = existingContent.replace(
+          /^(\s*)/gm,
+          `$1${LINE_COMMENT_START} `
+        );
+        cm.replaceRange(updatedContent, newRangeFrom, newRangeTo);
+      } else {
+        cm.replaceRange(
+          `${LINE_COMMENT_START} ${existingContent}`,
+          newRangeFrom,
+          newRangeTo
```
askoufis, line 322: Looks like these `replaceRange` calls only differ in there first parameter, so they can all probably be collapsed into a single call outside this `if else`.

askoufis (approved):

PR #326, opened 2024-02-22
Prevent new Playroom tabs from loading previously used titles

felixhabib:
Prevent new Playroom tabs with no set title from loading a recently used title.
New Playroom tabs should now have no set title by default.


src/StoreContext/StoreContext.tsx
michaeltaranto, line 1: Probably dont want to continue [setting the value] in storage either.

[setting the value]: https://github.com/seek-oss/playroom/pull/326/files#diff-23e36401909ea51a761508c39fe3466911be92d94143b0649494f4b642ff3136R391

askoufis (approved):

PR #327, opened 2024-02-22
Remove unnecessary "waits" in Cypress tests

felixhabib:
Cypress no longer requires "wait" usage for most situations. This change removes it for as many cases as possible, making tests run much faster.

### DIFF
--- a/cypress/support/utils.js
+++ b/cypress/support/utils.js
@@ -26,7 +26,10 @@ export const visit = (url) =>
       );
     });
 
-export const typeCode = (code, { delay = 200 } = {}) =>
+export const typeCode = (code) =>
+  getCodeEditor().focused().type(code, { force: true });
+
+export const typeCodeWithDelay = (code, { delay = 200 } = {}) =>
   getCodeEditor()
     .focused()
     .type(code, { force: true, delay })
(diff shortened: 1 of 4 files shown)


cypress/support/utils.js
```diff
@@ -26,7 +26,10 @@ export const visit = (url) =>
       );
     });
 
-export const typeCode = (code, { delay = 200 } = {}) =>
+export const typeCode = (code) =>
+  getCodeEditor().focused().type(code, { force: true });
+
+export const typeCodeWithDelay = (code, { delay = 200 } = {}) =>
```
askoufis, line 32: Is the typing delay the thing causing this function to make some tests work correctly, or is it the `wait` at the end, or both? Feel like a typing delay of 200ms per key stroke is way too much.

felixhabib: Looks like the delay is causing the tests to work. I've experimented with a few different values and looks like the `autocompletes` test in `editor.cy.js` fails with any nice round value lower than 100. The other tests that use it aren't as sensitive so I could just pass 100 as the delay value for that test only and have the default be lower but not sure what is cleaner. What are your thoughts?

askoufis: Seems like there's still some test flakiness, so maybe keep it at 200ms I guess? But we probably don't need the `WAIT_FOR_FRAME_TO_RENDER` at the end.

I might experiment with removing `force` as that could be getting in the way of things.

PR #328, opened 2024-02-23
Remove cypress delays, modify assertions to leverage cypress' retryability

askoufis:
Converting some of `then`s into [`should`](https://docs.cypress.io/api/commands/should)s and utilizing [`its`](https://docs.cypress.io/api/commands/its#Assertions) lets us lean into cypress' retry mechanisms, so we don't need to wait for iframe loads or add tons of delays during interactions anymore.

This blog post came in handy for the iframe stuff: https://www.cypress.io/blog/2020/02/12/working-with-iframes-in-cypress

The only exception is the `autocompletes` test as that relies on the autocomplete options to have rendered before the `enter` key is pressed.

Also bumped cypress to v13. Would be good to run CI a few times to confirm that it's stable enough for our liking.

### DIFF
--- a/cypress/e2e/editor.cy.js
+++ b/cypress/e2e/editor.cy.js
@@ -21,7 +21,7 @@ describe('Editor', () => {
   });
 
   it('autocompletes', () => {
-    typeCode('<F{enter} c{enter}={downarrow}{enter} />');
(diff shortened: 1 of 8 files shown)


felixhabib (approved):
🔥

PR #335, opened 2024-03-02
Remove dependency on `current-git-branch`

askoufis:
Replaces [`current-git-branch`](https://www.npmjs.com/package/current-git-branch), a relatively simple, yet likely unmaintained package, with a call to `git branch --show-current`.

### DIFF
--- a/lib/provideDefaultConfig.js
+++ b/lib/provideDefaultConfig.js
@@ -1,10 +1,21 @@
 const readPackage = require('read-pkg-up');
-const currentGitBranch = require('current-git-branch');
+const { execSync } = require('child_process');
+
+/**
+ * @returns {string | null} The current git branch name, or null if no branch is found
+ */
+const getGitBranch = () => {
+  try {
+    return execSync('git branch --show-current').toString().trim();
+  } catch (e) {
+    return null;
+  }
+};
 
 const generateStorageKey = () => {
   const pkg = readPackage.sync();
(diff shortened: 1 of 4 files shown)


lib/provideDefaultConfig.js
```diff
@@ -1,10 +1,21 @@
 const readPackage = require('read-pkg-up');
-const currentGitBranch = require('current-git-branch');
+const { execSync } = require('child_process');
+
+/**
+ * @returns {string | null} The current git branch name, or null if no branch is found
+ */
+const getGitBranch = () => {
+  try {
+    return execSync('git branch --show-current').toString().trim();
```
jahredhope, line 9: I've previously been burnt by OS differences when running raw child_process commands.

Is this something we should worry about?

I believe current-git-branch was using [execa](https://www.npmjs.com/package/execa) which builds in some "Make work on Windows" stuff.

askoufis: I think all of [the stuff `execa` (via `cross-spawn`) solves as far as Windows is concerned](https://github.com/moxystudio/node-cross-spawn#why) is irrelevant for this very simple command. I've just tested this on my Windows machine in both Cmd and PowerShell, and it worked fine.

jahredhope (approved):
Love the removal of the dep, and since it was tested in Windows seems good to me.

PR #336, opened 2024-03-02
Replace `query-string` dependency with `URLSearchParams`

askoufis:
The platform-native `URLSearchParams` has broad browser support, so we don't really have a need for `query-string` anymore.

### DIFF
--- a/src/Playroom/Frame.tsx
+++ b/src/Playroom/Frame.tsx
@@ -19,11 +19,15 @@ export default function Frame({
   components,
   FrameComponent,
 }: FrameProps) {
-  const { themeName, code } = useParams((rawParams) => ({
-    themeName:
-      typeof rawParams.themeName === 'string' ? rawParams.themeName : '',
-    code: typeof rawParams.code === 'string' ? rawParams.code : '',
-  }));
+  const { themeName, code } = useParams((rawParams) => {
+    const rawThemeName = rawParams.get('themeName');
+    const rawCode = rawParams.get('code');
+
+    return {
+      themeName: rawThemeName || '',
+      code: rawCode || '',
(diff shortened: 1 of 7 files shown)


src/Playroom/Frame.tsx
```diff
@@ -19,11 +19,15 @@ export default function Frame({
...
-  }));
+  const { themeName, code } = useParams((rawParams) => {
+    const rawThemeName = rawParams.get('themeName');
+    const rawCode = rawParams.get('code');
+
+    return {
+      themeName: rawThemeName || '',
+      code: rawCode || '',
+    };
+  });
```
askoufis, line 30: `query-string`'s API could return an array, but [`URLSearchParams` `.get` API](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get) returns `string | null`, so there's no need for a `typeof` check here anymore.

jahredhope (approved):
But what about my 8 year old Samsung phone that I refuse to update since I first bought it?

src/utils/params.ts
```diff
@@ -24,18 +21,18 @@ export function updateUrlCode(code: string) {
...
   try {
-    return queryString.parse(
+    return new URLSearchParams(
       playroomConfig.paramType === 'hash'
         ? location.hash.replace(/^#/, '')
         : location.search
     );
   } catch (err) {
-    return {};
+    return new URLSearchParams();
```
askoufis, line 30: This `try catch` might not even be needed anymore since AFAICT URLSearchParams constructor never throws. I couldn't make it throw on any input, and a cursory glance at the spec didn't mention throwing. But probably better to be safe than sorry.

PR #337, opened 2024-03-02
Upgrade `webpack-dev-server` to v5

askoufis:
None of the breaking changes in v5 affected our usage. See [migration guide](https://github.com/webpack/webpack-dev-server/blob/master/migration-v5.md).


jahredhope (approved):
👍 Node Engine field is already > 18.12 so makes sense to me

PR #340, opened 2024-03-06
Refactor logic to determine comment type for "Toggle comment" command

felixhabib:
Fixes an issue where full line expression slots outside any JSX tags would be line commented rather than block commented.

# Example of problem
https://github.com/seek-oss/playroom/assets/33821218/211da86d-31cc-438e-a823-645295b18daa

### DIFF
--- a/src/Playroom/CodeEditor/keymaps/comment.ts
+++ b/src/Playroom/CodeEditor/keymaps/comment.ts
@@ -243,6 +243,7 @@ const determineCommentType = (
     cm.getLine(to.line).trimEnd().endsWith(BLOCK_COMMENT_END);
 
   if (isFullExpressionSlot(lineTokens)) {
+  if (!isBlockComment && isFullExpressionSlot(lineTokens)) {
     return isJavaScriptMode ? 'block' : 'line';
   }
 
--- a/src/Playroom/CodeEditor/keymaps/comment.ts
+++ b/src/Playroom/CodeEditor/keymaps/comment.ts
@@ -198,6 +198,18 @@ function getUpdatedContent(existingContent: string, range: TagRange) {
 
 type CommentType = 'line' | 'block';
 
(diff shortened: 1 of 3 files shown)


src/Playroom/CodeEditor/keymaps/comment.ts
```diff
@@ -198,6 +198,19 @@ function getUpdatedContent(existingContent: string, range: TagRange) {
...
+const isOnlyWhitespace = (input: string) => /^\s+$/.test(input);
+
+const isFullExpressionSlot = (tokens: CodeMirror.Token[]) => {
+  const formattedLineTokens = tokens.filter(
+    (token) => token.type !== 'comment' && !isOnlyWhitespace(token.string)
+  );
+
+  return (
+    formattedLineTokens.at(0)?.string === '{' &&
+    formattedLineTokens.at(-1)?.string === '}'
```
michaeltaranto, line 210: Am I right that this is only intended to work on one line?

For example:
```ts
{() => {}
}
<div>First line</div>
<div>Second line</div>
<div>Third line</div>
```

Commenting out line 1 of this will result in (renders the closing curly in the frame):
```ts
{/* {() => {} */}
}
<div>First line</div>
<div>Second line</div>
<div>Third line</div>
```

Thoughts?

benjervis: I don't think there's any way to comment out line 1 that doesn't result in a syntax error, and I don't think the plugin should go beyond your selection to try and make your comment non-breaking. You just messed up the comment, at that point

felixhabib: Interesting, yeah it's only meant to work on the lines selected so in this case it's working as intended. I agree with Ben's point although it's annoying to get the closing curly rendered 😅

michaeltaranto (approved):

PR #350, opened 2024-05-10
Fix side panel exit animation

felixhabib:
Improve the side panel exit animation by not the hiding of panel content before the animation has completely finished. As a result, the panel will always display the content of the most recently active tab, even during the closing animation.

Using `react-transition-group` library to unmount the side panel on exit.

# Before (10% speed)

https://github.com/seek-oss/playroom/assets/33821218/fae1d41c-0949-46ef-aa3c-71f91e67009f

# After (10% speed)

https://github.com/seek-oss/playroom/assets/33821218/7f18d8b6-a0bf-4f99-8654-14bb964ef84c


cypress/support/utils.js
```diff
@@ -153,6 +153,9 @@ export const assertCodePaneLineCount = (lines) => {
   getCodeEditor().within(() =>
     cy.get('.CodeMirror-line').should('have.length', lines)
   );
+
+  // Wait after check to ensure original focus is restored
+  cy.wait(500); // eslint-disable-line @finsit/cypress/no-unnecessary-waiting
```
askoufis, line 158: Is this needed for every call of `assertCodePaneLineCount`, or does it only affect a specific one?

src/Playroom/Toolbar/Toolbar.css.ts
```diff
@@ -92,24 +92,52 @@ export const panel = style([
...
 
 export const preference = sprinkles({
   position: 'absolute',
   inset: 0,
 });
+
+export const transitionStyles = {
+  enter: style({
+    opacity: 0,
+    transform: `translateX(${calc(`${toolbarOpenSize}px`).multiply(0.3)})`,
```
askoufis, line 112: Worth bringing this out into a constant so we can name it and avoid the duplication.

.changeset/beige-tips-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+'playroom': patch
+---
+
+Fix minor issue in the side panel exit animation, preventing the contents from vanishing abruptly.
```
askoufis, line 5: Not sure if you were aiming to describe the problem or the result, but you kind of ended up with both?
```suggestion
Fixes a bug in the side panel exit animation that was causing the contents to vanish abruptly
```

src/Playroom/Toolbar/Toolbar.tsx
```diff
@@ -148,58 +154,65 @@ export default ({ themes: allThemes, widths: allWidths, snippets }: Props) => {
...
+          timeout={ANIMATION_TIMEOUT}
+          classNames={styles.transitionStyles}
+          mountOnEnter
+          unmountOnExit
+          onEnter={() => setLastActivePanel(activeToolbarPanel)}
+          onExited={() => setLastActivePanel(undefined)}
+        >
+          <div className={styles.panel} id="custom-id">
+            {lastActivePanel === 'snippets' && (
+              <div className={styles.preference}>
```
askoufis, line 168: All of the conditionally rendered panels share a top-level `div` with just a `styles.preference` classname. Could the `div` maybe be hoisted outside the conditional render, and hence deduplicated, or maybe conditionally applied alongside `styles.panel` on line 166?

cypress/support/utils.js
felixhabib: It's needed for at least a couple and it's used 10 times total. The tests aren't always consistent but I'll make it an optional prop anyway to be more explicit when the wait is needed.

src/Playroom/Toolbar/Toolbar.css.ts
felixhabib: I've refactored slightly, but I think it might be better to keep the full styles here so that it's easier to compare the differences in each transition stage, and looks like the more common way this library is used. Thoughts?

askoufis: Oh, I was thinking extracting the entire string into a constant, but I think using `translateX(30%)` is better than referencing `toolbarOpenSize`. Nice.

src/Playroom/Toolbar/Toolbar.tsx
felixhabib: Good idea, I reckon just wrapping it around all the conditional renders is probably the simplest option.

src/Playroom/Toolbar/Toolbar.tsx
```diff
@@ -148,58 +160,56 @@ export default ({ themes: allThemes, widths: allWidths, snippets }: Props) => {
...
+        <CSSTransition
+          in={isOpen}
+          timeout={ANIMATION_TIMEOUT}
+          classNames={styles.transitionStyles}
+          mountOnEnter
+          unmountOnExit
+          onExited={() => setLastActivePanel(undefined)}
+        >
+          <div className={styles.panel} id="custom-id">
+            <div className={styles.preference}>
```
askoufis, line 172: These can be merged now that the inner `div` isn't doing anything other than applying a style.

In fact, is it worth just merging the `panel` style with the `preference` style?

PR #353, opened 2024-05-22
Update Cypress tests to reference elements by name and role instead of `data-testid`

felixhabib:
* Use Cypress Testing Library to query elements by name and role.
* Convert cypress e2e test files from Javascript to Typescript for better type safety in tests.
* Add aria-labels to certain elements for improved accessibility and as a testing element reference.
* Remove all unused `data-testid` attributes from elements.

Still using `data-testid` for the splashscreen, preview frames and preview frame names. Should find appropriate accessibility labels and use that for a testing reference in a seperate PR.


cypress/support/utils.ts
```diff
@@ -221,20 +195,14 @@ export const replaceInCode = (term, replaceWith) => {
   }
 };
 
-/**
- * @param {number} line
- */
-export const jumpToLine = (line) => {
+export const jumpToLine = (line: number | string) => {
```
michaeltaranto, line 198: Is there a use case for `string` here? Feel like all line numbers should be `number`

cypress/support/utils.ts
```diff
@@ -149,26 +133,26 @@ export const assertCodePaneContains = (text) => {
...
 
-export const assertCodePaneLineCount = (lines, { wait } = {}) => {
+export const assertCodePaneLineCount = (lines: number, wait?: number) => {
   getCodeEditor().within(() =>
     cy.get('.CodeMirror-line').should('have.length', lines)
   );
 
   // Wait after check to ensure original focus is restored
-  if (typeof wait === 'number') {
+  if (wait) {
```
michaeltaranto, line 142: This will have a behaviour change for zero. Not sure if `cy.wait(0)` pushes to next tick or not.

cypress/support/utils.ts
```diff
@@ -17,71 +18,65 @@ export const getPreviewFrames = () => cy.get('[data-testid="previewFrame"]');
...
-export const filterSnippets = (search) => {
-  cy.get('[data-testid="filterSnippets"]').type(search);
+export const filterSnippets = (search: string) => {
+  cy.findByRole('searchbox', { name: 'Search snippets' }).type(search);
 };
 
-export const assertSnippetsListIsVisible = () =>
-  cy.get('[data-testid="snippets"]').should('be.visible');
+export const assertSnippetsSearchFieldIsVisible = () =>
+  cy.findByRole('searchbox', { name: 'Search snippets' }).should('be.visible');
```
michaeltaranto, line 53: ❤️

cypress/support/utils.ts
```diff
@@ -17,71 +18,65 @@ export const getPreviewFrames = () => cy.get('[data-testid="previewFrame"]');
...
-export const selectWidthPreferenceByIndex = (index) =>
-  cy
-    .get('[data-testid="toggleFrames"]')
-    .then((el) => el.get(0).click())
-    .get('[data-testid="widthsPreferences"] label')
-    .eq(index)
-    .then((el) => el.get(0).click());
+export const selectWidthPreference = (width: number) => {
+  cy.findByRole('button', { name: 'Configure visible frames' }).click();
+  // Force needed as the checkbox is covered by 'Checkmark' svg
```
michaeltaranto, line 31: Can we remove the force and turn pointer events off on the svg?

src/Playroom/Snippets/Snippets.tsx
```diff
@@ -159,13 +160,12 @@ export default ({ snippets, onHighlight, onClose }: Props) => {
...
             }
           }}
-          data-testid="filterSnippets"
         />
       </div>
       <ul
         className={styles.snippetsContainer}
-        data-testid="snippet-list"
         ref={listEl}
+        aria-label="Filtered snippets list"
```
askoufis, line 168: Not sure if there are conventions for labeling lists, but given that the element is a `ul`, adding `list` to the label itself seems a bit redundant.

```suggestion
        aria-label="Filtered snippets"
```

cypress/e2e/toolbar.cy.ts
```diff
@@ -17,7 +17,13 @@ describe('Toolbar', () => {
     const widthIndexToSelect = 1;
 
     assertFramesMatch(frames);
-    selectWidthPreferenceByIndex(widthIndexToSelect);
+
+    const widthSizeToSelectAsNumber = parseInt(
+      frames[widthIndexToSelect].replace('px', ''),
+      10
+    );
+
```
michaeltaranto, line 25: Given we convert back to a string in `selectWidthPreference` I feel like this parsing is not necessary.

cypress/support/utils.ts
felixhabib: We let you give a line number and a character number as a string and we've got one test that does that (e.g. "6:10"). Not sure if it'd be better to give `character: number` as a seperate argument?

michaeltaranto: I would suggest separate utils, breaking out something like `jumpToPosition({ line: number, ch: number })`

cypress/support/utils.ts
felixhabib: Good point. Thinking now, it's probably better to always have the wait as `500` anyway, since that's the only wait time we use

src/Playroom/FramesPanel/FramesPanel.tsx
```diff
@@ -160,17 +159,16 @@ export default ({ availableWidths, availableThemes }: FramesPanelProps) => {
                 }}
               />
             ))}
-          </div>
+          </Stack>
 
           {hasThemes ? (
-            <div data-testid="themePreferences">
+            <Stack space="none">
```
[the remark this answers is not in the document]
felixhabib: `div` works here, I just found that using `Stack` makes it clearer why there is a wrapper element here, particularly on line 136. Thoughts?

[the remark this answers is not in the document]
askoufis: Fair enough, I guess as long as it's communicating that this is a vertical stack of elements it makes sense. 👍

cypress/e2e/toolbar.cy.ts
felixhabib: I think it makes sense to have the `width` type as `number` in `selectWidthPreference` to make it clearer that it shouldn't get the 'px' string. Although it is a bit awkward to use here with how this test is set up. Thoughts?

michaeltaranto: I think we can agree that this logic shouldnt be in the test, so then the question becomes where should it be.

The config for `widths` takes an array of numbers, its in the Frames component that we add the `px` suffix.

So maybe we change the above `frames` array to an array of numbers, and we append the `px` in `assertFramesMatch`? That way the api for that util accepts the same structure as the playroom config

Thoughts?

felixhabib: Ahh yes good idea. I've moved the formatting logic into the `assertFramesMatch` function.

cypress/support/utils.ts
```diff
@@ -149,26 +134,38 @@ export const assertCodePaneContains = (text) => {
...
   }
 };
 
-export const assertFramesMatch = (matches) =>
+export const assertFramesMatch = (
+  frames: number[] | Array<[frameTheme: string, frameWidth: number]>
+) => {
+  const formattedFrames = frames.map((match) =>
+    typeof match === 'number' ? `${match}px` : `${match[0]} – ${match[1]}px`
+  );
```
askoufis, line 156:
```suggestion
  const formattedFrames = frames.map((frame) =>
    typeof frame === 'number' ? `${frame}px` : `${frame[0]} – ${frame[1]}px`
  );
```

askoufis (approved):

PR #355, opened 2024-05-23
Update `Stack` component to use CSS `gap`

felixhabib:
* Update `Stack` component to use CSS `gap`, simplifying the component
* Remove unused `Divider` component
* Callout browser policy in README


askoufis (approved):
I _think_ this should have a changeset. It's something that benefits the UI, so should be released. Don't know how to word it though, as it's not really worth detailing implementation changes. 🤔

askoufis:
Branch is not up to date, but you can still merge for some reason.
image

I think that should be disabled in the repo settings though 🤔

.changeset/khaki-seahorses-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+'playroom': minor
+---
+
+Optimise Playroom UI performance.
```
askoufis, line 5: I wouldn't really call out this as a perf change. Maybe it _is_ just worth saying something like `Use CSS gap for spacing` or something.

I _do_ think it's worth adding a note to the README saying that Playroom will only ever utilize features supported by all browers, but it assumes you have the latest stable version of any of the major browsers.

askoufis (approved):

PR #357, opened 2024-06-03
Update inline spacing treatment, using modern CSS features

felixhabib:
Taking advantage of modern CSS spacing features for a better dev experience. This change:

* Adds a `small` size to the spacing scale (`6px`) and increases `medium` spacing to `8px`
* Converts all existing uses of `medium` spacing to `small` spacing
* Updates the `Inline` component to:
  * use gap for spacing,
  * use the space scale from `sprinkles`, and
  * include an `alignY` prop
* Replaces uses of horizontal padding with flex gap on inline elements (using the `Inline` component or CSS where appropriate)
* Uses a 3 column grid for keyboard shortcut key arrangement
* Replaces vertical padding in Frame Panel with `Stack` spacing

Using the changeset from https://github.com/seek-oss/playroom/pull/355 as this work is related.


src/Playroom/Frames/Frames.css.ts
```diff
@@ -4,26 +4,21 @@ import { sprinkles } from '../sprinkles.css';
...
-    // // Simulate centering when fewer frames than viewport width.
-    '::before': {
-      content: '""',
-      flex: 1,
-    },
-    '::after': {
-      content: '""',
-      flex: 1,
-    },
+    marginInline: 'auto',
```
askoufis, line 21: 💅

src/Playroom/SettingsPanel/SettingsPanel.tsx
```diff
@@ -71,12 +71,14 @@ const KeyboardShortcut = ({
...
+    <Inline space="small" alignY="center">
+      <div className={styles.keyboardShortcutTitle}>
+        <Text>{description}</Text>
+      </div>
       <Text size={isMac() ? 'large' : 'standard'}>
-        <Inline space="xxsmall">{shortcutSegments}</Inline>
+        <div className={styles.keyboardShortcutKeys}>{shortcutSegments}</div>
       </Text>
-    </div>
+    </Inline>
```
askoufis, line 81: My bad if we already discussed this, but why use grid here instead of sticking with `Inline`?

src/Playroom/FramesPanel/FramesPanel.css.ts
```diff
@@ -48,15 +45,8 @@ export const reset = style([
 
 export const label = style([
```
askoufis, line 46: This doesn't need to be wrapped in a `style` call anymore.

askoufis (approved):

PR #358, opened 2024-06-03
Fix iframe re-loading when serving certain fixtures

askoufis:
When serving either the `basic` or the `typescript` fixture locally (via `build:basic && serve:basic` or `build:typescript && serve:typescript`), each iframe would re-load on every code change. This shouldn't happen when using `paramType: 'hash'` because hash replacements typically don't cause page navigation.

It turns out that the culprit was the default behaviour of the `serve` CLI. By default, [`cleanUrls` is `true`](https://github.com/vercel/serve-handler?tab=readme-ov-file#cleanurls-booleanarray). This results in all requests for `.html` files being redirected to the same path but without `.html` (see screenshot). This was occurring for iframe updates, which point directly to `/frame.html`, resulting in a navigation and full re-load of the frame.

image

The fix is to exclude `/frame.html` from this behaviour by configuring this inside a `serve.json` file in each fixture directory.

It turns out the `themed` fixture already had a `serve.json` configuration, but it was just blanket-disabling this behaviour for all paths, which doesn't really achieve anything, but it means that you have to explicitly navigate to `/index.html` to get the `themed` fixture to load locally, which is annoying.

> [!NOTE]
> If you test this locally, you might still see the full re-load happening. This is probably due to your browser cache. Disabling it for the page, or forcing a full page reload should fix it.


felixhabib (approved):
🚀

PR #359, opened 2024-06-05
Add `Box` component

felixhabib:
This change:

* Moves `vars` into its own file (from `sprinkles.css.ts`)
* Adds a `Box` component, getting props from `sprinkles`
* Adds `auto` as a `margin` space option in `sprinkles`
* Uses `Box` in place of `flexGrow` and `margin: auto` for layout where appropriate
* Removes unused classes in [Toolbar.css.ts](https://github.com/seek-oss/playroom/pull/359/files#diff-c87c0765a8d411b8fef9f7ed8329e04305b71aacf45709a158c28a7eefc1fb0f)

Continuing the work from https://github.com/seek-oss/playroom/pull/357 and https://github.com/seek-oss/playroom/pull/355 so will share the same changeset


src/Playroom/Toolbar/Toolbar.tsx
```diff
@@ -89,7 +89,7 @@ export default ({ themes: allThemes, widths: allWidths, snippets }: Props) => {
       )}
       <div className={styles.sidebar}>
         <div className={styles.buttons}>
-          <div className={styles.topButtons}>
+          <div>
```
askoufis, line 92: Should this be a `Box` with `transition="medium"` to replace the deleted class?

src/Playroom/Frames/Frames.css.ts
askoufis, line 1: Looks like the `root` class can also have the `style` call removed.

src/Playroom/Box/Box.tsx
```diff
@@ -0,0 +1,33 @@
...
+  component = 'div',
+  className,
+  ...restProps
+}: BoxProps) => {
+  const atomProps: Record<string, unknown> = {};
+
+  for (const key in restProps) {
+    if (sprinkles.properties.has(key as keyof Sprinkles)) {
+      atomProps[key] = restProps[key as keyof typeof restProps];
+      delete restProps[key as keyof typeof restProps];
```
askoufis, line 25: While this differs from the Braid implementation, I suppose there's no need to create a new object for the native props since `restProps` is already its own object. Nice.

askoufis (approved):

PR #360, opened 2024-08-30
Apply `title` from url on page load

michaeltaranto:
Previously the document `title` would only update when the frames panel is open.
The title is now correctly reflected from the url on page load.


felixhabib (approved):
🙌

PR #361, opened 2024-08-30
start: Disable webpack error overlay

michaeltaranto:
Prevent the default webpack dev server error overlay from blocking the preview frames in `start` mode.
Playroom handles it's own errors, and this would block the preview frames and need to be dismissed manually.

### DIFF
--- a/.changeset/rich-geese-confess.md
+++ b/.changeset/rich-geese-confess.md
@@ -1,0 +1,8 @@
+---
+'playroom': patch
+---
+
+start: Disable webpack error overlay
+
+Prevent the default webpack dev server error overlay from blocking the preview frames in `start` mode.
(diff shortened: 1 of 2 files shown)


.changeset/rich-geese-confess.md
```diff
@@ -0,0 +1,8 @@
+---
+'playroom': patch
+---
+
+start: Disable webpack error overlay
+
+Prevent the default webpack dev server error overlay from blocking the preview frames in `start` mode.
+Playroom handles it's own errors, and this would block the preview frames and need to be dismissed manually.
```
felixhabib, line 8:
```suggestion
Playroom handles its own errors, and this would block the preview frames and need to be dismissed manually.
```

felixhabib (approved):
Works for me, awesome!

PR #363, opened 2024-09-09
Save code editor size preference as percentage

felixhabib:
See changeset.

Setting the defaults to 40% as these look pretty good even at smaller screen sizes, but open to suggestions if this should be different.

### DIFF
--- a/.changeset/clever-keys-act.md
+++ b/.changeset/clever-keys-act.md
@@ -1,0 +1,6 @@
+---
+'playroom': minor
+---
+
+Save editor height and width preferences as a percentage of the viewport size, rather than a fixed pixel value.
+This prevents the editor from obscuring preview panels when toggling the browser tools on/off.
--- a/.changeset/clever-keys-act.md
+++ b/.changeset/clever-keys-act.md
@@ -3,4 +3,4 @@
 ---
 
 Save editor height and width preferences as a percentage of the viewport size, rather than a fixed pixel value.
-This prevents the editor from obscuring preview panels when toggling the browser tools on/off.
+This prevents the editor from obscuring preview panels when toggling the browser tools on/off or resizing the window.
--- a/src/Playroom/Playroom.tsx
+++ b/src/Playroom/Playroom.tsx
@@ -21,7 +21,7 @@ import * as styles from './Playroom.css';
 import { Box } from './Box/Box';
 
 import { assignInlineVars } from '@vanilla-extract/dynamic';
-import { editorSize as editorSize } from './Playroom.css';
+import { editorSize } from './Playroom.css';
 
 const resizableConfig = (position: EditorPosition = 'bottom') => ({
   top: position === 'bottom',
@@ -135,7 +135,7 @@ export default ({ components, themes, widths, snippets }: PlayroomProps) => {
   const isVerticalEditor = editorPosition === 'right';
   const isHorizontalEditor = editorPosition === 'bottom';
   const sizeStyles = {
-    height: isHorizontalEditor ? editorHeight : 'auto', // issue in ff & safari when not a string
+    height: isHorizontalEditor ? editorHeight : 'auto',
     width: isVerticalEditor ? editorWidth : 'auto',
   };
   const editorContainer =
@@ -176,7 +176,6 @@ export default ({ components, themes, widths, snippets }: PlayroomProps) => {
         </Helmet>
       )}
       <Box
-        id="preview"
         className={[
           styles.previewContainer,
           editorHidden
--- a/src/Playroom/Playroom.tsx
+++ b/src/Playroom/Playroom.tsx
@@ -7,7 +7,7 @@ import Frames from './Frames/Frames';
 import { WindowPortal } from './WindowPortal';
 import type { Snippets } from '../../utils';
 import componentsToHints from '../utils/componentsToHints';
-import Toolbar, { toolbarItemCount } from './Toolbar/Toolbar';
+import Toolbar from './Toolbar/Toolbar';
 import ChevronIcon from './icons/ChevronIcon';
 import { StatusMessage } from './StatusMessage/StatusMessage';
 import {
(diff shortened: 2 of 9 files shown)


src/StoreContext/StoreContext.tsx
```diff
@@ -328,21 +328,25 @@ const createReducer =
 
       case 'updateEditorHeight': {
         const { size } = action.payload;
-        store.setItem('editorHeight', size);
+        const viewportHeight = window.innerHeight;
+        const heightPercentage = (size / viewportHeight) * 100;
```
michaeltaranto, line 332: Thoughts on rounding this number?

Wondering whether we should format it as a percentage here too, given there are no calculations on it. Means we dont need to template it everywhere we use it.

src/Playroom/Playroom.tsx
```diff
@@ -133,11 +133,18 @@ export default ({ components, themes, widths, snippets }: PlayroomProps) => {
     </Fragment>
   );
 
+  if (editorHeightPercentage < 0 || editorHeightPercentage > 100) {
+    throw new Error('editorHeightPercentage must be a value between 0 and 100');
+  }
+  if (editorWidthPercentage < 0 || editorWidthPercentage > 100) {
+    throw new Error('editorWidthPercentage must be a value between 0 and 100');
+  }
```
michaeltaranto, line 141: If we are defaulting it when would this be needed? Only asking because we dont validate other properties this way, and if we [format the stored value to a percentage](https://github.com/seek-oss/playroom/pull/363/files#r1751023795) then this will require parsing.

felixhabib: Ahh good point. I can't see it causing an issue, so probably not necessary. Have removed

src/StoreContext/StoreContext.tsx
```diff
@@ -508,8 +515,10 @@ export const StoreProvider = ({
           payload: {
             ...(code ? { code } : {}),
             ...(editorPosition ? { editorPosition } : {}),
-            ...(editorHeight ? { editorHeight } : {}),
-            ...(editorWidth ? { editorWidth } : {}),
+            ...(editorHeightPercentage
+              ? { editorHeightPercentage }
+              : undefined),
+            ...(editorWidthPercentage ? { editorWidthPercentage } : undefined),
```
michaeltaranto, line 521: Consistency nitpick 😄 

```suggestion
            ...(editorHeightPercentage ? { editorHeightPercentage } : {}),
            ...(editorWidthPercentage ? { editorWidthPercentage } : {}),
```

.changeset/clever-keys-act.md
```diff
@@ -0,0 +1,6 @@
+---
+'playroom': minor
+---
+
+Save editor height and width preferences as a percentage of the viewport size, rather than a fixed pixel value.
+This prevents the editor from obscuring preview panels when toggling the browser tools on/off.
```
michaeltaranto, line 6:
```suggestion
This prevents the editor from obscuring preview panels when toggling the browser tools on/off or resizing the window.
```

src/StoreContext/StoreContext.tsx
```diff
@@ -490,8 +497,8 @@ export const StoreProvider = ({
       ]) => {
         const code = codeFromQuery || storedCode || exampleCode;
         const editorPosition = storedPosition;
-        const editorHeight = storedHeight;
-        const editorWidth = storedWidth;
```
michaeltaranto, line 494: Should we add any backwards compatibility here? If the stored value was not a percentage, convert it to a percentage as part of the initial load?

If not, what is the UX? Does the user need to resize the editor to get the new behaviour?

felixhabib: Fair enough, I've updated it to have backwards compatibility

src/StoreContext/StoreContext.tsx
```diff
@@ -490,17 +514,28 @@ export const StoreProvider = ({
...
+
+        const editorHeight =
+          (typeof storedHeight === 'string' ? storedHeight : null) ||
+          (typeof storedHeight === 'number' ? `${storedHeight}px` : null) ||
+          defaultEditorSize;
```
michaeltaranto, line 521: These could be simplified further, if its a number convert to pixel, otherwise assume percentage, if falsey fallback to default.
```suggestion
          (typeof storedHeight === 'number' ? `${storedHeight}px` : storedHeight) ||
          defaultEditorSize;
```

src/StoreContext/StoreContext.tsx
```diff
@@ -490,17 +514,28 @@ export const StoreProvider = ({
...
+
+        const editorWidth =
+          (typeof storedWidth === 'string' ? storedWidth : null) ||
+          (typeof storedWidth === 'number' ? `${storedWidth}px` : null) ||
+          defaultEditorSize;
```
michaeltaranto, line 526: As above
```suggestion
        const editorWidth =
          (typeof storedWidth === 'number' ? `${storedWidth}px` : storedWidth) ||
          defaultEditorSize;
```

src/Playroom/Playroom.tsx
```diff
@@ -136,8 +135,8 @@ export default ({ components, themes, widths, snippets }: PlayroomProps) => {
   const isVerticalEditor = editorPosition === 'right';
   const isHorizontalEditor = editorPosition === 'bottom';
   const sizeStyles = {
-    height: isHorizontalEditor ? `${editorHeight}px` : 'auto', // issue in ff & safari when not a string
-    width: isVerticalEditor ? `${editorWidth}px` : 'auto',
+    height: isHorizontalEditor ? editorHeight : 'auto', // issue in ff & safari when not a string
```
michaeltaranto, line 138: If the height is always a string now given its a percentage, we can remove this comment:
```suggestion
    height: isHorizontalEditor ? editorHeight : 'auto',
```

src/Playroom/Playroom.tsx
```diff
@@ -7,22 +7,21 @@ import Frames from './Frames/Frames';
...
-
 import { CodeEditor } from './CodeEditor/CodeEditor';
 
 import * as styles from './Playroom.css';
-import { toolbarOpenSize } from './Toolbar/Toolbar.css';
-import { toolbarItemSize } from './ToolbarItem/ToolbarItem.css';
+import { Box } from './Box/Box';
+
+import { assignInlineVars } from '@vanilla-extract/dynamic';
+import { editorSize as editorSize } from './Playroom.css';
```
michaeltaranto, line 24:
```suggestion
import { editorSize } from './Playroom.css';
```

askoufis (approved):
This is great! Much nicer experience when resizing.

michaeltaranto (approved):
So good. Love where this landed, very tidy 👏

PR #364, opened 2024-09-09
Improve snippets scrolling behaviour

felixhabib:
See changeset.

While the animations look nice, they feel sluggish and can get slow when you start to select items not currently visible on screen.


src/Playroom/Snippets/Snippets.tsx
```diff
@@ -94,15 +49,21 @@ export default ({ snippets, onHighlight, onClose }: Props) => {
...
+
   const filteredSnippets = useMemo(
     () => filterSnippetsForTerm(snippets, searchTerm),
     [searchTerm, snippets]
   );
 
+  const highlightedItem =
+    typeof highlightedIndex === 'number'
+      ? document.getElementById(
+          `${filteredSnippets[highlightedIndex].group}_${filteredSnippets[highlightedIndex].name}_${highlightedIndex}`
```
michaeltaranto, line 61: Given the formatting here, maybe its useful to have a function for generating the ID for a snippet? That way the formatting between here and the list item ID remain in sync and linked in code.

src/Playroom/Snippets/Snippets.tsx
```diff
@@ -173,22 +131,27 @@ export default ({ snippets, onHighlight, onClose }: Props) => {
...
-                  <Strong>{snippet.group}</Strong>
-                  <span className={styles.snippetName}>{snippet.name}</span>
-                </Text>
-              </span>
+              <div
+                className={classnames(styles.snippet, {
+                  [styles.highlight]: isHighlighted,
+                })}
+              >
+                <span style={{ display: 'block', position: 'relative' }}>
```
michaeltaranto, line 148: Wonder if we need both elements here. Is there anyway to flatten this down the `li` only? Or at least combine this `div` and `span`

src/Playroom/Snippets/Snippets.css.ts
```diff
@@ -48,6 +52,8 @@ export const snippet = style([
     paddingX: 'xlarge',
   }),
   {
+    marginInline: snippetsBorderSpace,
```
felixhabib, line 55: Setting the margin on the individual snippets rather than the container to avoid margin being applied before the actual scrollbar. This is an issue in master

src/Playroom/Snippets/Snippets.tsx
```diff
@@ -94,15 +52,27 @@ export default ({ snippets, onHighlight, onClose }: Props) => {
...
+      ? getSnippetId(
+          filteredSnippets[highlightedIndex].group,
+          filteredSnippets[highlightedIndex].name,
+          highlightedIndex
+        )
+      : null;
+
+  const highlightedItem = highlightedItemId
+    ? document.getElementById(highlightedItemId)
+    : null;
```
felixhabib, line 72: Maybe this could be more elegant, but I'm not sure the best way

michaeltaranto: Yeah agree we could refine this a bit. My suggestion would be for the `getSnippetId` function to take a `Snippet` so something like:
```ts
function getIdForSnippet(snippet: Snippet, index: number) {
  return `${snippet.group}_${snippet.name}_${index}`;
}
```
then the above could be more like:
```ts
if (typeof highlightedIndex === 'number') {
  const highlightedItem = document.getElementById(
    getIdForSnippet(filteredSnippets[highlightedIndex], index)
  );

  highlightedItem?.scrollIntoView({ block: 'nearest' });
}
```

michaeltaranto (approved):
Like where this landed. Great stuff

PR #365, opened 2024-09-12
Add shorthand for `start`, `build` and `serve` scripts without theme variant

felixhabib:
For easier local dev muscle memory


package.json
```diff
@@ -19,6 +19,9 @@
     "cypress": "start-server-and-test build-and-serve:all '9000|9001|9002' 'cypress run'",
     "cypress:dev": "start-server-and-test start:all '9000|9001|9002' 'cypress open --browser chrome --e2e'",
     "cypress:verify": "cypress verify",
+    "start": "pnpm start:basic",
+    "build": "pnpm build:basic",
+    "serve": "pnpm serve:basic",
```
michaeltaranto, line 24: Im not as sure about these two as i am about `start`. For example, running `build` on the repo might produce an unexpected outcome.

felixhabib: Fair enough, I only care about start really

michaeltaranto (approved):

PR #369, opened 2024-10-08
Save editor hidden status to Playroom url

felixhabib:
See changeset

Last feature I want to get in before cutting a release

### DIFF
--- a/cypress/e2e/urlHandling.cy.ts
+++ b/cypress/e2e/urlHandling.cy.ts
@@ -42,6 +42,7 @@ describe('URL handling', () => {
       /*
       The CodeMirror element is not visible, but it is in the DOM
       This test fails because the element doesn't meet Cypress's requirements for being hidden
+      Not sure why Cypress's hidden requirement isn't met
       https://docs.cypress.io/guides/core-concepts/interacting-with-elements#An-element-is-considered-hidden-if
       */
       // cy.get('.CodeMirror-code').should('be.hidden');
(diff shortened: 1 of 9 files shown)


cypress/e2e/urlHandling.cy.ts
```diff
@@ -30,6 +30,22 @@ describe('URL handling', () => {
...
+      cy.get('textarea').should('not.be.focused');
+
+      // Todo - write a test that checks the CodeMirror element is not visible
+      /*
+      The CodeMirror element is not visible, but it is in the DOM
+      This test fails because the element doesn't meet Cypress's requirements for being hidden
+      https://docs.cypress.io/guides/core-concepts/interacting-with-elements#An-element-is-considered-hidden-if
+      */
+      // cy.get('.CodeMirror-code').should('be.hidden');
+    });
```
felixhabib, line 48: This is proving very difficult to test. I don't think it is worth holding up the release unless there is an easy fix

cypress/e2e/urlHandling.cy.ts
```diff
@@ -30,6 +30,22 @@ describe('URL handling', () => {
...
     });
+
+    it('editor hidden', () => {
+      cy.visit(
+        'http://localhost:9000/#?code=N4IgpgJglgLg9gJwBJQhMA7EAuGCCuYAvkA'
+      );
+
+      cy.get('textarea').should('not.be.focused');
+
+      // Todo - write a test that checks the CodeMirror element is not visible
```
askoufis, line 41: Feel like this todo is redundant given the explanation below.

cypress/e2e/urlHandling.cy.ts
```diff
@@ -30,6 +30,22 @@ describe('URL handling', () => {
...
+      /*
+      The CodeMirror element is not visible, but it is in the DOM
+      This test fails because the element doesn't meet Cypress's requirements for being hidden
```
askoufis, line 44: Could be worth clarifying that we don't know what's causing it to fail the hidden requirement either.

askoufis (approved):

cypress/e2e/urlHandling.cy.ts
felixhabib: I think I'll keep the todo only but remove the surrounding comment, so when a fix is found it prompts us to change it for both tests

Issue #368, opened 2024-10-07
Duplicated components in viewport but not in code panel 😣


yceballost:
I guys, I found something weird.

When you select snippet in a fast way, the snippet is placed in the code panel but in the viewport you can see this snippet duplicated.

A video to show this behavior. This video is made with **Braid playroom**. I found it in our playroom implementation but I confirm that is happening in other playrooms. 

https://github.com/user-attachments/assets/32cd5887-3ca4-423b-8dc4-52be084f61e0

In the video I was using Arc as a browser (in incognito mode to avoid problems with extensions), but also is happening in Safari

felixhabib:
Thanks for the issue! Playing with this on a couple machines and browsers, I'm unable to replicate this behaviour. Does this occur consistently for you, and only when selecting snippets with the cursor? I appreciate any additional info.

yceballost:
Yes, only when I open snippet panel with cursor and select a snippet "quite" fast. 

My steps to try to replicate is:
1. Put a snippet
2. Delete this snippet
3. Open snippet panel with cursor and select one (fast)

I tried the same in a Macbook Pro with Chrome and Sonoma (14.6.1) and I also can replicate the issue.

felixhabib:
Thanks for this! I was able to replicate the issue and have found a fix

yceballost:
thanks for the quick fix!! 🥰🥰

PR #371, opened 2024-10-10
fix: Disable snippet previews while snippet panel is closing

felixhabib:
Resolves https://github.com/seek-oss/playroom/issues/368

This issue occurs when the user moves their cursor over a snippet in the snippet panel, while the panel is closing.
This triggers the frames to enter a mode for previewing snippets, which is only ended when the snippets panel is initially closed.

This change prevents snippets from being previewed after the panel begins to close.

### DIFF
--- a/.changeset/four-peaches-attack.md
+++ b/.changeset/four-peaches-attack.md
@@ -1,0 +1,8 @@
+---
+'playroom': patch
+---
+
+Disable snippet previews while the snippets panel is closing.
+
+Previously, snippet previews could be triggered while the snippet panel was closing, causing preview frames to enter an invalid state.
+Previewing a snippet will now only work when the snippet panel is open.
--- a/src/Playroom/Snippets/Snippets.tsx
+++ b/src/Playroom/Snippets/Snippets.tsx
@@ -13,6 +13,7 @@ import * as styles from './Snippets.css';
 type HighlightIndex = number | null;
 type ReturnedSnippet = Snippet | null;
 interface Props {
+  isOpen: boolean;
   snippets: PlayroomProps['snippets'];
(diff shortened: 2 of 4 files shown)


.changeset/four-peaches-attack.md
```diff
@@ -0,0 +1,8 @@
+---
+'playroom': patch
+---
+
+Disable snippet previews while the snippets panel is closing.
+
+Previously, snippet previews could be triggered while the snippet panel was closing, causing preview frames to enter an invalid state.
+Previewing a snippet will now only work when the snippet panel is open.
```
askoufis, line 8: This feels a bit too much like implementation detail to me. I'd just mention the bug and say it's fixed.

```suggestion
Fixes a bug that was causing duplicate snippet previews and broken preview updates when a snippet was clicked and the mouse was moved while the snippets panel was closing
```

It's a bit of a long sentence, but I think it gets the point across.

src/Playroom/Snippets/Snippets.tsx
```diff
@@ -142,9 +143,15 @@ export default ({ snippets, onHighlight, onClose }: Props) => {
...
+                isOpen && !isHighlighted
+                  ? () => {
+                      setHighlightedIndex(index);
+                    }
+                  : undefined
               }
-              onMouseDown={() => closeHandler(filteredSnippets[index])}
+              onMouseDown={() => {
+                closeHandler(filteredSnippets[index]);
+              }}
```
askoufis, line 154: Could make this diff smaller by removing the curly braces from the functions. Github wouldn't let me suggest a change :(

michaeltaranto (approved):
Nice one!

askoufis (approved):

PR #381, opened 2024-12-10
Test `componentToHints`, exclude test files from be published

askoufis:
I have a branch to remove `lodash` as a dep, which is used by `componentToHints`, so I wanted to add some tests to ensure nothing breaks.

Also noticed that test files are being published, so I've added a `.npmignore` to prevent that (while also renaming all `.spec` tiles to `.test`).


src/Playroom/Playroom.tsx
```diff
@@ -59,7 +61,7 @@ const getTitle = (title: string | undefined) => {
 };
 
 export interface PlayroomProps {
-  components: Record<string, ComponentType>;
+  components: Record<string, ComponentType<any>>;
```
askoufis, line 64: `ComponentType` defaults to a component with no props, specifying `any` here permits components with or without props, which is obviously more accurate than before.

src/index.d.ts
```diff
@@ -32,4 +32,7 @@ interface Window {
 }
 
 declare const __PLAYROOM_GLOBAL__CONFIG__: InternalPlayroomConfig;
-declare const __PLAYROOM_GLOBAL__STATIC_TYPES__: any;
+declare const __PLAYROOM_GLOBAL__STATIC_TYPES__: Record<
+  string,
+  Record<string, unknown>
+>;
```
askoufis, line 38: This type isn't as precise as it could be. I'm unsure of the exact type it _should_ be, but it's better than `any`.

src/utils/componentsToHints.ts
```diff
@@ -3,9 +3,10 @@ import omit from 'lodash/omit';
 import parsePropTypes from 'parse-prop-types';
 import type { PlayroomProps } from '../Playroom/Playroom';
 
-const staticTypes = __PLAYROOM_GLOBAL__STATIC_TYPES__;
```
askoufis, line 6: Moving this into `Playroom.tsx` makes this function much easier to test as there's no need to modify globals in the tests.

src/.npmignore
```diff
@@ -0,0 +1,2 @@
+*.test.*
+
```
askoufis, line 2: Tested this with `npm pack`:

Before: 128 files being packaged.
After: 124 files being packaged.

There are exactly 4 test files inside `src`, so this seems correct.

src/utils/componentsToHints.test.ts
```diff
@@ -0,0 +1,70 @@
+import componentsToHints from './componentsToHints';
+// @ts-expect-error
+import * as PropTypeComponents from '../../cypress/projects/themed/components';
+import * as TypeScriptComponents from '../../cypress/projects/typescript/components';
+
+describe('componentsToHints', () => {
+  it('should support javascript components with proptypes', () => {
+    const result = componentsToHints({
+      Foo: PropTypeComponents.Foo,
+      Bar: PropTypeComponents.Bar,
```
felixhabib, line 10:
```suggestion
      Bar: PropTypeComponents.Bar,
      Foo: PropTypeComponents.Foo,
```
Ditto for second test

felixhabib (approved):

src/index.d.ts
askoufis: I've made it more precise in https://github.com/seek-oss/playroom/pull/381/commits/7354aee2c8e8fa87bebaa211352c4bd40847fc00. This won't change anything at runtime of course, so the types aren't really that important, but if we ever convert more stuff to typescript it could be useful.

PR #388, opened 2025-01-03
Set default colour scheme to 'system'

felixhabib:


michaeltaranto (commented):
What are your thoughts on making the default configurable via the config?

felixhabib:
@michaeltaranto possibly, not sure how much this would be changed? This feels like a sensible default given the user can give a preference

PR #390, opened 2025-02-07
feat: React 19 support

imjordanxd:
### Overview
Adds support for React 19. Comments added where appropriate.

Closes #389

PR #391, opened 2025-02-14
chore: Update way to check react version to switch to use react-dom/client

vctqs1:


src/Playroom/Toolbar/Toolbar.tsx
```diff
@@ -157,13 +158,14 @@ export default ({ themes: allThemes, widths: allWidths, snippets }: Props) => {
         </div>
         <CSSTransition
           in={isOpen}
+          nodeRef={nodeRef}
```
vctqs1, line 161: To do this you dont need to use `ReactDOM.findDOMNode` which is removed in react@19
https://github.com/reactjs/react-transition-group/blob/2989b5b87b4b4d1001f21c8efa503049ffb4fe8d/src/ReplaceTransition.js#L32-L35

vctqs1:
Hi @felixhabib, I'm creating this to enhance its capacity with React/ReactDOM@19. Could you please have a look at this? 

The change in`CSSTransition` doesn't affect any behavior, but it's another way to avoid using `ReactDOM.findDOMNode`, which is being removed in React@19.
Here is the documentation for react-transition-group on using nodeRef:
🔗 [React Transition Group - CSS Transition](https://reactcommunity.org/react-transition-group/css-transition)

And here are the specific lines mentioning it:
🔗 [GitHub - ReplaceTransition.js#L32-L35](https://github.com/reactjs/react-transition-group/blob/2989b5b87b4b4d1001f21c8efa503049ffb4fe8d/src/ReplaceTransition.js#L32-L35)

With these small updates, I can put react@19 in resolution in my project without any breaking changes.

felixhabib:
Thanks a lot for the work! We've added React 19 support in https://github.com/seek-oss/playroom/pull/390 which covers similar changes, but appreciate the PR nontheless 🙏

askoufis:
@vctqs1 [`v0.40.0`](https://github.com/seek-oss/playroom/releases/tag/v0.40.0) has been released. It includes React 19 support.

PR #392, opened 2025-03-12
Enable embedded CSS formatting on save

askoufis:
Adding the `postcss` plugin to playroom's prettier config allows formatting of CSS inside `style` tags with a `jsx` attribute.

I knew about formatting inside tagged template literals, e.g. `css`, so initially this PR documented a method of adding a `css` identity function to your Playroom scope in order to enable this.

However, while investigating the effort required to make a prettier plugin to enable formatting CSS without the need for a `css` tagged template literal, I ended up looking at [the code for formatting embedded CSS within the prettier repo](https://github.com/prettier/prettier/blob/6aa8f9725a436d4815b7ceb87391952e8a06fa6a/src/language-js/embed/css.js#L67). This lead me to find out that just adding a `jsx` attribute to the `style` tag _also_ enables formatting of CSS.

AFAIK this isn't properly documented anywhere, other than inside the prettier source code. Found [this old feature request](https://github.com/prettier/prettier/issues/1948) where it was mentioned that this would be useful for various CSS-in-JS libraries.

[Surge example](https://playroom--ea9bff9f31a659c2f433ba8acd936312450cf065.surge.sh/?code=N4Igxg9gJgpiBcIA8BnALgTwDYwAQCsUAPAPmAAMALNAWy12FwCMIAnWV%2BXARgAcjcKCFgCWUXKxjiAvuWlIA9OmwwSIADQg0lGDRgoEAbS069AeQB2cALqaA7mO0H4hgMwAmAAzXpQA). If surge is broken, test out following code in a fixture:

```jsx
<style jsx>{`html { border: 1px solid red }`}</style>
```

### DIFF
--- a/cypress/e2e/editor.cy.ts
+++ b/cypress/e2e/editor.cy.ts
@@ -36,7 +36,7 @@ describe('Editor', () => {
 
   it('formats css in a style block', () => {
     typeCode(
-      '<style jsx>{{}`html {{} border: 1px solid red; {}}`{rightarrow}{backspace}{}}'
+      '<style jsx>{{}`html {{} border: 1px solid red; {}}{rightarrow}{}}'
     );
     assertCodePaneLineCount(1);
     formatCode();
--- a/cypress/e2e/editor.cy.ts
+++ b/cypress/e2e/editor.cy.ts
@@ -1,3 +1,4 @@
+import dedent from 'dedent';
 import {
   formatCode,
   typeCode,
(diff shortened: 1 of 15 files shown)


cypress/e2e/editor.cy.ts
```diff
@@ -32,4 +33,19 @@ describe('Editor', () => {
     formatCode();
     assertCodePaneLineCount(6);
   });
+
+  it.only('formats css in a style block', () => {
```
felixhabib, line 37:
```suggestion
  it('formats css in a style block', () => {
```

askoufis:
@felixhabib Turns out you can just add a `jsx` attribute to a `style` tag and it will format the CSS inside. Found it by digging in to [the prettier source code](https://github.com/prettier/prettier/blob/6aa8f9725a436d4815b7ceb87391952e8a06fa6a/src/language-js/embed/css.js#L67).

michaeltaranto (approved):
Be great to be able to opt in by default for this, but its a clean step forward 👏

PR #396, opened 2025-03-14
Drop support for React 17

askoufis:
Dropping support for React 17. This lets us simplify `renderElement` and remove a hack from our webpack config.

Additionally, noticed that installing the latest version of Playroom in a repo that installs `@types/react@18` results in type conflicts because Playroom's now on `@types/react@19`. Expanding both `@types/react` and `@types/react-dom` to include v18 should address this.


michaeltaranto:
Nice!! There was some other [webpack hackery] to enable 17 and 18 support that we should be able to clean up too.

[webpack hackery]: https://github.com/seek-oss/playroom/blob/75a73531c3633c92aa5c4da26d7f3d9d6bd9526d/lib/makeWebpackConfig.js#L199-L200

michaeltaranto (approved):

PR #401, opened 2025-03-23
Update PNPM to v10, remove `volta` config, use node 20

askoufis:
Lockfile format changed, hence the big diff. `onlyBuiltDependencies` is also now necessary as we have a few deps that run a build as a `postinstall` step.

Finally, removed the `volta` config in favour of the existing `.nvmrc` as we're moving away from `volta` internally, and most tooling supports `.nvmrc` for declaring the node version.


michaeltaranto (commented):
Thoughts on updating node while you're at it too?

askoufis:
> Thoughts on updating node while you're at it too?

Sure. I guess we could go to 20 for now since 18 is EOL soon.

michaeltaranto (approved):

PR #410, opened 2025-05-12
Improve editor animation, prevent searching editor when hidden

felixhabib:
Breaking off this change before other layout changes, mainly bringing the toolbar out of the editor

### DIFF
--- a/.changeset/spicy-ants-push.md
+++ b/.changeset/spicy-ants-push.md
@@ -4,4 +4,5 @@
 
 Refactor layout.
 
-Improve the animation for the toggling code editor. Prevent code contents from being searchable when the editor is hidden.
+Improve the code editor show/hide animation.
+Prevent code contents from being searchable when the editor is hidden.
--- a/.changeset/spicy-ants-push.md
+++ b/.changeset/spicy-ants-push.md
@@ -1,0 +1,7 @@
+---
+'playroom': minor
+---
+
+Refactor layout.
+
+Improve the animation for the toggling code editor. Prevent code contents from being searchable when the editor is hidden.
--- a/cypress/e2e/urlHandling.cy.ts
+++ b/cypress/e2e/urlHandling.cy.ts
@@ -37,15 +37,7 @@ describe('URL handling', () => {
       );
 
       cy.get('textarea').should('not.be.focused');
-
-      // Todo - write a test that checks the CodeMirror element is not visible
-      /*
-      The CodeMirror element is not visible, but it is in the DOM
-      This test fails because the element doesn't meet Cypress's requirements for being hidden
-      Not sure why Cypress's hidden requirement isn't met
-      https://docs.cypress.io/guides/core-concepts/interacting-with-elements#An-element-is-considered-hidden-if
-      */
-      // cy.get('.CodeMirror-code').should('be.hidden');
+      cy.get('.CodeMirror-code').should('be.hidden');
     });
   });
 
@@ -86,8 +78,7 @@ describe('URL handling', () => {
       );
 
       cy.get('textarea').should('not.be.focused');
-
-      // Todo - write a test that checks the CodeMirror element is not visible
+      cy.get('.CodeMirror-code').should('be.hidden');
     });
   });
 });
--- a/src/Playroom/Playroom.css.ts
+++ b/src/Playroom/Playroom.css.ts
@@ -65,14 +65,10 @@ export const resizable = style([
 // before applying minimum and maximum sizes
 export const resizableSize = styleVariants({
   right: {
-    width: `clamp(${[`${MIN_WIDTH}px`, editorSize, MAX_WIDTH].join(
-      ', '
-    )}) !important`,
+    width: `clamp(${MIN_WIDTH}px, ${editorSize}, ${MAX_WIDTH}) !important`,
   },
   bottom: {
-    height: `clamp(${[`${MIN_HEIGHT}px`, editorSize, MAX_HEIGHT].join(
-      ', '
-    )}) !important`,
+    height: `clamp(${MIN_HEIGHT}px, ${editorSize}, ${MAX_HEIGHT}}) !important`,
   },
 });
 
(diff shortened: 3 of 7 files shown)


cypress/e2e/urlHandling.cy.ts
```diff
@@ -37,15 +37,7 @@ describe('URL handling', () => {
...
       cy.get('textarea').should('not.be.focused');
-
-      // Todo - write a test that checks the CodeMirror element is not visible
-      /*
-      The CodeMirror element is not visible, but it is in the DOM
-      This test fails because the element doesn't meet Cypress's requirements for being hidden
-      Not sure why Cypress's hidden requirement isn't met
-      https://docs.cypress.io/guides/core-concepts/interacting-with-elements#An-element-is-considered-hidden-if
-      */
-      // cy.get('.CodeMirror-code').should('be.hidden');
```
felixhabib, line 48: This layout change fixes this test

src/Playroom/Playroom.css.ts
```diff
@@ -27,71 +34,60 @@ globalStyle('body', {
...
   },
-]);
+  bottom: {
+    maxHeight: MAX_HEIGHT,
+  },
+});
+
+export const resizableUnavailable = style({
+  transition: `width ${ANIMATION_TIMEOUT}ms ease, height ${ANIMATION_TIMEOUT}ms ease`,
+});
```
askoufis, line 90: IMO longhand properties would be a bit more readable here, but I guess slightly more CSS. Up to you if you want to swap.
```suggestion
export const resizableUnavailable = style({
  transitionProperty: 'width, height',
  transitionDuration: `${ANIMATION_TIMEOUT}ms`,
  transitionTimingFunction: 'ease',
});
```

src/Playroom/Playroom.tsx
```diff
@@ -116,7 +138,7 @@ export default ({ components, themes, widths, snippets }: PlayroomProps) => {
 
   const codeEditor = (
     <Fragment>
-      <div className={styles.editorContainer}>
+      <div inert={editorHidden} className={styles.editorContainer}>
```
askoufis, line 141: Ooooo, nice. Didn't know about this attribute. Very handy.

src/Playroom/Playroom.css.ts
```diff
@@ -27,71 +34,60 @@ globalStyle('body', {
...
-      },
-    },
+// override re-resizable's inline style
+// clamp ensures editor transitioning to correct size
+// before applying minimum and maximum sizes
+export const resizableSize = styleVariants({
+  right: {
+    width: `clamp(${[`${MIN_WIDTH}px`, editorSize, MAX_WIDTH].join(
+      ', '
+    )}) !important`,
```
askoufis, line 70: I feel like _not_ using `Array.join` here would look cleaner. Just comma-separated values.

```suggestion
  right: {
    width: `clamp(${MIN_WIDTH}px, ${editorSize}, ${MAX_WIDTH}) !important`,
```

.changeset/spicy-ants-push.md
```diff
@@ -0,0 +1,7 @@
+---
+'playroom': minor
+---
+
+Refactor layout.
+
+Improve the animation for the toggling code editor. Prevent code contents from being searchable when the editor is hidden.
```
askoufis, line 7:
```suggestion
Improve the code editor show/hide animation. Prevent code contents from being searchable when the editor is hidden.
```

src/Playroom/Playroom.css.ts
askoufis, line 70: Would it be worth documenting why `!important` is necessary here? E.g. what style are you fighting that makes it necessary, so that if that style were to disappear one day then so could this `!important`.

askoufis (commented):
Very slick! Just a few comments.

src/Playroom/Playroom.css.ts
felixhabib: How is this for a new comment above the class? I'd like it to be more succinct if it can be
```
/*
Min/max values are disabled during transitions to allow zero height.
This class ensures the open editor transition uses the correct size,
rather than `editorSize`, before min/max values are applied.

Resizable's `size` prop would instead transition to `editorSize`,
so `!important` is used to override this.
*/
```

askoufis: Looks good 👍

askoufis (approved):

src/Playroom/Playroom.css.ts
```diff
@@ -27,71 +34,58 @@ globalStyle('body', {
...
+// override re-resizable's inline style
+// clamp ensures editor transitioning to correct size
+// before applying minimum and maximum sizes
+export const resizableSize = styleVariants({
+  right: {
+    width: `clamp(${MIN_WIDTH}px, ${editorSize}, ${MAX_WIDTH}) !important`,
   },
-]);
+  bottom: {
+    height: `clamp(${MIN_HEIGHT}px, ${editorSize}, ${MAX_HEIGHT}}) !important`,
```
michaeltaranto, line 71: This rule is broken with the extra brace in it, but wondering whether we need these clamps.

Can we add a `minWidth` to the `resizeableAvailable` rule below and then have these just set their value as `editorSize`?

```suggestion
    height: `clamp(${MIN_HEIGHT}px, ${editorSize}, ${MAX_HEIGHT}) !important`,
```

michaeltaranto: In fact, playing with it now I dont think we need this rule at all anymore.

felixhabib: The issue should be apparent when size percentage is less than the minimum size. You can see it if you set the editor size to the minimum and then shrink the viewport size to be very small

src/Playroom/Playroom.css.ts
```diff
@@ -7,11 +7,18 @@ import {
...
+import {
+  ANIMATION_TIMEOUT,
+  toolbarItemCount,
+  toolbarOpenSize,
+} from './constants';
 
 export const MIN_HEIGHT = toolbarItemSize * toolbarItemCount;
 export const MIN_WIDTH = toolbarOpenSize + toolbarItemSize + 80;
 
+const MAX_HEIGHT = '90vh';
```
michaeltaranto, line 19: Even though these values already existed, this should probably be reduced to `80vh`. At `90vh` the frames have already disappeared entirely.

michaeltaranto (approved):

PR #417, opened 2025-05-29
Improved handling of rendering errors

michaeltaranto:
Errors occurring during render no longer replace the frame contents with a red error message.
Instead, the error is caught and overlaid on top of the last successful render result (when possible).


src/Playroom/vars.css.ts
```diff
@@ -1,7 +1,7 @@
 import { createGlobalTheme } from '@vanilla-extract/css';
 
-export const fontFamily = 'Helvetica, arial, sans-serif';
-export const vars = createGlobalTheme(':root', {
+const fontFamily = 'Helvetica, arial, sans-serif';
+export const rawValues = {
```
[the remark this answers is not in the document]
michaeltaranto: No as its in a child document in the iframe. But im looking at a refactor atm to see if i can get the error message outside of the frame to work.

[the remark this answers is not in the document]
askoufis: Ah, didn't realize. Error message outside the frame sounds like a good move 👍.

src/Playroom/Frame/frameMessaging.tsx
```diff
@@ -0,0 +1,41 @@
...
+}) => {
+  useEffect(() => {
+    window.parent.postMessage({
+      source: PlayroomErrorSource,
+      message: errorMessage,
+    });
+  }, [errorMessage]);
+
+  return null;
+};
```
askoufis, line 20: Could this be a hook? Should it be a hook? E.g. `useRegisterErrorSender`.

michaeltaranto: Its used as a component because its passed to the `Frame` component, and that behaviour is shared across the preview entrypoint too (and that passes more than just a Hook)

src/Playroom/Frame/Frame.tsx
```diff
@@ -0,0 +1,53 @@
...
+  themeName,
+  theme,
+  FrameComponent,
+  ErrorComponent,
+}: FrameProps) {
+  const [error, setError] = useState('');
+
+  return (
+    <ErrorBoundary
+      fallbackRender={() => <ErrorMessage errorMessage={error} />}
```
felixhabib, line 38: Wonder if `ErrorMessage` can be unified with `ErrorComponent` here

cypress/e2e/renderingErrors.cy.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    typeCode('<div style={{}{{} background {}}{}}>', 150);
+
+    assertFirstFrameError('background is not defined');
+  });
+
+  it('show error when component does not exist', () => {
+    loadPlayroom();
+
+    typeCode('<Fake />', 150);
```
michaeltaranto, line 26:
```suggestion
    typeCode('<div style={{}{{} background {}}{}}>');

    assertFirstFrameError('background is not defined');
  });

  it('show error when component does not exist', () => {
    loadPlayroom();

    typeCode('<Fake />');
```

felixhabib (approved):

PR #418, opened 2025-06-03
Migrate some internals to typescript

felixhabib:
These changes are based on the work from #313 (tysm @TheMightyPenguin 🙏). Splitting out this work into multiple new PRs for ease of reviewing and to be based off fresher code.

This first PR focusses on migrating the top level `src` files to Typescript, and any dependant files where required.

**NB**: The `loadPlayroom` function will now clear code immediately after loading if no `initialCode` argument is provided. This is to ensure saved code from `indexedDB` doesn't conflict with test expectations.


askoufis (dismissed):

src/userDefinedModules.d.ts
michaeltaranto, line 1: Lets rename this to `playroomModuleAliases.d.ts`. The "user defined" prefix threw me for a second, prefer the name to more clearly define its usage.

src/Playroom/RenderCode/RenderCode.js
```diff
@@ -1,8 +1,7 @@
-// eslint-disable-next-line import-x/no-unresolved
-import useScope from '__PLAYROOM_ALIAS__USE_SCOPE__';
 import React, { createElement, Fragment } from 'react';
 import scopeEval from 'scope-eval';
 
+import useScope from '../../useScope';
```
askoufis, line 4: Turns out the `useScope` entrypoint was never actually used at all, but now it is.

michaeltaranto (approved):

PR #421, opened 2025-06-11
Convert `utils` to TypeScript, bundle them for CJS and ESM

askoufis:
`utils/index.ts` contains public utility functions. In order to convert them to TypeScript without shipping TypeScript code (bad), we need to bundle it.

Have been experimenting with `tsdown` recently, and it seems stable and featureful enough to suitable for this. Happy to use the more stable `tsup` if necessary. Both tools handle ESM/CJS builds as well as package exports definitions, which is all we really need.


package.json
```diff
@@ -2,18 +2,27 @@
...
+  },
   "files": [
     "CHANGELOG.md",
     ".babelrc",
     "images",
     "lib",
     "src",
-    "utils"
+    "utils",
+    "dist"
```
askoufis, line 25: Functions in `utils` are still reference internally via relative paths, rather than `import { createUrl } from 'playroom'`, so we still need to include the `utils` directory in the published package.

This may change in the future if (for example) we end up splitting out these utils into a separate package (or the playroom UI itself moves to a separate package.

utils/index.ts
```diff
@@ -1,13 +1,32 @@
...
-const compressParams = ({
+import type { Widths } from '../src/Playroom/Playroom';
+
+export interface Snippet {
+  group: string;
+  name: string;
+  code: string;
+}
+
+export type Snippets = Snippet[];
```
michaeltaranto, line 11: Is this deliberate? If so, should we add a changeset for exposing snippet types?

askoufis: All these types are already exposed in the old `index.d.ts` ([braid usage](https://github.com/seek-oss/braid-design-system/blob/eb5a8cbdc3d7ff5be9d5056784e4ba933a143848/packages/braid-design-system/src/lib/components/private/Snippets.ts#L2)). I think if you don't specify `exports` explicitly, all types in a `.d.ts` file are exported. The new bundle _does_ explicitly export though.

michaeltaranto (approved):

PR #439, opened 2025-07-22
Migrate top-level layout to CSS Grid

michaeltaranto:
Refactor the top-level layout of Playroom to use CSS Grid in preparation for upcoming UI features.

Additionally, migrate away from `re-resizable` package for panel resizing in favour of a custom implementation to improve UI performance.

### DIFF
--- a/src/components/CodeEditor/CodeEditor.tsx
+++ b/src/components/CodeEditor/CodeEditor.tsx
@@ -73,6 +73,7 @@ export const CodeEditor = ({
   onChange,
   previewCode,
 }: Props) => {
+  const mounted = useRef(false);
   const editorInstanceRef = useRef<Editor | null>(null);
   const insertionPointRef = useRef<ReturnType<Editor['addLineClass']> | null>(
     null
@@ -161,9 +162,10 @@ export const CodeEditor = ({
   useEffect(() => {
     if (editorInstanceRef.current) {
       if (
-        editorInstanceRef.current.hasFocus() ||
-        code === editorInstanceRef.current.getValue() ||
-        previewCode
+        mounted.current &&
+        (editorInstanceRef.current.hasFocus() ||
+          code === editorInstanceRef.current.getValue() ||
+          previewCode)
       ) {
         return;
       }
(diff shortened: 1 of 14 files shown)


src/components/Playroom/ResizeHandle.tsx
```diff
@@ -0,0 +1,100 @@
...
+      const movePosition = resolvePosition(moveEvent, pagePos);
+      const delta = movePosition - startPosition;
+      onResize(startSize - delta * modifier);
+    };
+
+    const stopHandler = () => {
+      if (typeof onResizeEnd === 'function') {
+        const endSize = ref.current?.[elementSize] ?? 0;
+        onResizeEnd(endSize);
+      }
```
michaeltaranto, line 67:
```suggestion
      const endSize = ref.current?.[elementSize] ?? 0;
      onResizeEnd?.(endSize);
```

src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -220,6 +215,10 @@ export const CodeEditor = ({
         if (!editorHidden) {
           setCursorPosition(cursorPosition);
         }
+        // Set mounted flag on next tick
+        setTimeout(() => {
+          mounted.current = true;
+        }, 1);
```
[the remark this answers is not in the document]
michaeltaranto: It is required unfortunately. The old resize library was rendering its children (the code editor) later, which impacted when the editor instance was created and provided with content.

Now that we are rendering the editor sooner, there are behaviours wired up to the editor instance (cursor position and change handling) that have conditional logic based on focus (which it receives sooner now).

I'll update the comment above to be less about "next tick" and more about following the initialisation process.

src/components/CodeEditor/CodeEditor.tsx
```diff
@@ -220,6 +215,10 @@ export const CodeEditor = ({
         if (!editorHidden) {
           setCursorPosition(cursorPosition);
         }
+        // Set mounted flag on next tick
```
michaeltaranto, line 218:
```suggestion
        /**
         * This workaround delays the setting of the mounted flag. It allows
         * the behaviours wired up via `useEffect` to complete without being
         * interrupted by change or focus events.
         */
```

askoufis (approved):

PR #446, opened 2025-08-28
Update method to manage tab titles

### LINKED ISSUE #445: React 19 support (peerDependency errors)
Hi again! I'm looking to update to React 19, however I'm getting a peerDependency warning via playroom > react-helmet -> react-side-effect.

The issue stems from `react-side-effect` library, which has not been updated to support React 19: https://github.com/gaearon/react-side-effect/issues/89

However, it comes to playroom via `react-helmet`, which itself has not been maintained for years now, see:
- https://github.com/nfl/react-helmet/issues/716
- https://github.com/nfl/react-helmet/issues/717

From those issues come a couple of react-helmet replacements, it seems that:
- `react-helmet-async` also has no support for React 19, however it has dropped react-side-effect dependency.
- `@dr.pogodin/react-helmet` only supports React 19, which is incompatible with playroom supporting both 18 and 19 currently.
 
If playroom dropped support for React 18 altogether, then the dep probably would not be needed at all, as it seems it's only used for updating `<title>`, which I believe React 19 has native support for (see https://react.dev/blog/2024/12/05/react-19#support-for-metadata-tags).

However, if you want to support both 18 and 19 for some time, not sure what the solution should be. 

I did find [react-helmet-pro](https://www.npmjs.com/package/react-helmet-pro), which seems like a recent fork that supports both 18 and 19, but it has basically no usage, so not sure if or how well it works as a replacement.

---------------------

Happy to contribute if we can agree on how to resolve this.

felixhabib:
Resolves #445 

When React 18 is no longer supported, we should be able to render the `title` tag ourselves and remove the hook


src/utils/useDocumentTitle.ts
```diff
@@ -0,0 +1,13 @@
+import { useEffect } from 'react';
+
+export function useDocumentTitle(newTitle?: string): void {
+  useEffect(() => {
+    if (typeof document === 'undefined') {
+      return;
+    }
```
michaeltaranto, line 7: This wont be necessary in an effect.

src/components/Playroom/Playroom.tsx
```diff
@@ -69,6 +69,9 @@ export default () => {
     },
     dispatch,
   ] = useContext(StoreContext);
+
+  useDocumentTitle(title === undefined ? undefined : getTitle(title));
```
jahredhope, line 73: `getTitle` isn't used elsewhere so it no-longer needs to accept `undefined` with this code.

But it then checks for truthy, which means it differs behaviour based only on "" (Line 40-47).

Is this intended? Otherwise can getTruthy be updated to just check for truthy/undefined and return undefined, and remove the `window?.__playroomConfig__.title` behaviour all together.

```suggestion
  useDocumentTitle(getTitle(title));
```

felixhabib: Thoughts on just ditching `getTitle` all together?

```tsx
const configTitle = window?.__playroomConfig__?.title || 'Playroom';
const documentTitle = title ? `${title} | ${configTitle}` : configTitle;

useDocumentTitle(documentTitle);
```

felixhabib: Tried to simplify another way. Let me know if you have feedback

src/utils/useDocumentTitle.ts
```diff
@@ -0,0 +1,36 @@
...
+
+interface UseDocumentTitleProps {
+  title?: string;
+  suffix?: string;
+}
+
+export function useDocumentTitle({
+  title,
+  suffix,
+}: UseDocumentTitleProps): void {
```
michaeltaranto, line 13: Declaring return types, particularly `void`, is not really a practice we have employed in the repo. Suggest keeping things consistent.

```suggestion
}: UseDocumentTitleProps) {
```

src/utils/useDocumentTitle.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export function useDocumentTitle({
+  title,
+  suffix,
+}: UseDocumentTitleProps) {
+  useEffect(() => {
+    if (title !== undefined) {
+      document.title = getTitle({ title, suffix });
+    }
```
michaeltaranto, line 17: This guard prevents the default case that we are replacing, where the fall through is the suffix.

felixhabib: Oh yikes, fixing

src/utils/useDocumentTitle.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  const configTitle = playroomConfig?.title;
+
+  if (configTitle) {
+    return `${configTitle} | ${suffix}`;
+  }
+
+  return suffix;
+}
```
michaeltaranto, line 31: Sorry, feel like this function can be simplified to make it clearer what is happening. Thoughts?

```suggestion
function getTitle({
  title,
  suffix = 'Playroom',
}: UseDocumentTitleProps): string {
  const resolvedTitle = title || playroomConfig?.title;
  if (resolvedTitle) {
    return `${resolvedTitle} | ${suffix}`;
  }
  
  return suffix;
}
```

felixhabib: Yep can be, thought that's a bit harder to parse all the cases but happy to change

michaeltaranto (approved):

PR #476, opened 2026-01-22
Prebundle src

mihkeleidast:
Currently, playroom ships the entire src directory as-is, meaning that the job of compiling playrooms internal components and styles, etc falls to the user. This also means that there are quite a few dependencies included just because of this, mainly a bunch of Vanilla Extract related deps.

What if playroom would prebundle its source code, compiling out the TypeScript and Vanilla Extract code, and leaving just normal CSS/JS to be imported by webpack?

This cleans up a bunch of code and deps - they can be devDeps now. Maybe some more can be move, e.g. some of the `@types/*` packages? The webpack config is quite complex, I'm not exactly sure how much of that was just needed to compile src previously.


package.json
```diff
@@ -80,21 +80,12 @@
     "@babel/parser": "^7.23.4",
     "@babel/preset-env": "^7.20.2",
     "@babel/preset-react": "^7.18.6",
```
mihkeleidast: @michaeltaranto I'd appreciate if you could take a hard look at the dependencies, if what I moved to devDeps makes sense, and if anything else could potentially be moved. A bit hard for me to judge, as I don't know what has been added and why.

Also the webpack config and loaders, which I trimmed down - maybe something can be simplified further there as well.

tsdown.src.config.ts
```diff
@@ -0,0 +1,31 @@
...
+  },
+  format: ['esm'],
+  dts: true,
+  clean: false,
+  outDir: 'dist/src',
+  exports: false,
+  platform: 'browser',
+  plugins: [vanillaExtractPlugin()],
+  // Doesn't affect the bundle but suppresses a warning we don't care about
+  copy: [{ from: 'images', to: 'dist' }],
```
michaeltaranto: Thoughts on copying over the [`template.html`] to the `app` directory? That way the `makeWebpackConfig` file only operates on the `dist/app` directory (can pull the resolution of this path into a variable to be shared between all references in that file).

[`template.html`]: https://github.com/seek-oss/playroom/pull/476/changes#diff-0036953330ca473a5195d429439f536e856bcc68e0fc7990c8857e2f6a63bfcaL166

tsdown.src.config.ts
```diff
@@ -0,0 +1,31 @@
...
+    'defaultModules/useScope': 'src/defaultModules/useScope.ts',
+  },
+  format: ['esm'],
+  dts: true,
+  clean: false,
+  outDir: 'dist/src',
+  exports: false,
+  platform: 'browser',
+  plugins: [vanillaExtractPlugin()],
+  // Doesn't affect the bundle but suppresses a warning we don't care about
```
michaeltaranto: This comment was originally against the `external`. We should shift it back down

package.json
michaeltaranto: I dont recall why we have a requirement for `@babel/core` (unless its a peer dep of something we use).

The `@babel/preset-react` package is [part of the default webpack config] we provide for consumers, so we should add that back in.

Can all the `@types/*` packages be moved to devdeps now?

[part of the default webpack config]: https://github.com/mihkeleidast/playroom/blob/e824cd9ff4443636c1b24adf475689fbffdb9a30/lib/makeDefaultWebpackConfig.js#L15

tsdown.app.config.ts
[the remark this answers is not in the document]
michaeltaranto: Im not sure i agree with the top-level `app` directory, feel like its clearer to have all compiled output under the parent `dist` directory

mihkeleidast:
@michaeltaranto anything I can do to push this forward?

michaeltaranto:
Sorry! The hold out on this one was the dev experience. When working on Playroom this serves the built artefacts only, rather than the source files being edited. I think that was the key blocker that I never got back to investigating.

package.json
```diff
@@ -51,6 +47,7 @@
     "version": "changeset version",
     "release": "pnpm package && changeset publish",
     "package": "tsdown",
+    "package:watch": "tsdown --watch --clean=false",
```
mihkeleidast: > Sorry! The hold out on this one was the dev experience. When working on Playroom this serves the built artefacts only, rather than the source files being edited. I think that was the key blocker that I never got back to investigating.

I added a `package:watch` script, which currently needs to be run in parallel to `start:basic` for example. We could also change the existing start:* scripts to start that before.

Speed-wise, I guess it's a little bit slower than before due to 2 separate bundlers needing to do some work sequentially, but not sure if it's a problem.

We could also do some trickery to use the source directly when running in dev mode in this repo, as it was before, but personally I think it's best to use the output from tsdown when developing, otherwise there might be some quirks when they don't compile exactly the same.

mihkeleidast: HMR did not work for me, but I checked out main branch and it did not work for me there either - not sure if it's supposed to work?

askoufis:
I've pushed [a commit](https://github.com/seek-oss/playroom/pull/476/commits/dc2419849791d419e91bffbf790526fa55f12345) that does a few things:
- Local dev now works. You should now be able to run `pnpm package:watch` and `pnpm start` and have changes to source code be picked up by the dev server.
- Emit `debug` VE identifiers when running `pnpm package:watch`, but `short` identifiers during `pnpm package`. 
- Replaced `exports` with `imports` for entrypoints. App entries and default modules are not public API, so `imports` is a better fit for them. The downside here is that tsdown won't manage them for you like it can for `exports`, but I think that's a worthwhile tradeoff to make.
- Consolidated image assets and the app template into a `static` directory that's copied into `dist` by `tsdown`.
- Added a small `Contributing` section at the bottom of the readme just to note down how to run local dev now that it's slightly more involved than before.
- Removed the v18 version range from the `@types/react` and `@types/react-dom` dev deps

Happy to tweak entrypoint naming and file locations some more, but I think everything works now.

lib/.npmignore
askoufis, line 1: No longer necessary now that `tsdown` is bundling everything and we no longer ship source code.

askoufis:
Changes since the previous comment:

- With some added import conditions, local dev now works the same as before. Just run `pnpm start`: no watched build required.
- Moved a few Vanilla Extract dependencies back out of dev deps as they are used by the playroom package. Note that these deps don't need a bundler because they're not generating CSS, they're just doing runtime stuff.
- Reworded some changesets

michaeltaranto (approved):

mihkeleidast:
Cheers, happy to see this finally merged 🎉

package.json
```diff
@@ -78,20 +112,11 @@
...
-    "@capsizecss/vanilla-extract": "^2.0.4",
     "@soda/friendly-errors-webpack-plugin": "^1.8.1",
-    "@types/codemirror": "^5.60.5",
-    "@types/react": "^18.0.0 || ^19.0.0",
-    "@types/react-dom": "^18.0.0 || ^19.0.0",
-    "@vanilla-extract/css": "^1.9.2",
     "@vanilla-extract/css-utils": "^0.1.3",
     "@vanilla-extract/dynamic": "^2.1.2",
-    "@vanilla-extract/integration": "^8.0.4",
     "@vanilla-extract/sprinkles": "^1.5.1",
```
mihkeleidast, line 119: Minor issue with keeping this in deps - sprinkles will still pull in `@vanilla-extract/css` via peerDeps, so users won't win as much... IIRC when this was in devDeps, playroom bundled the small runtime function needed from sprinkles, which would be less overhead.

Also not exactly sure if css-utils is necessary as a dep? It should be compiled away.

But overall this PR should still be quite a reduction in deps :)

askoufis: @mihkeleidast Good pickup. I misread [this error in CI](https://github.com/seek-oss/playroom/actions/runs/35553223361/job/106191754712#step:9:42) on a test branch and accidentally moved `css-utils` too. I think in newer `tsdown` versions these deps get silently but in the current version in the repo it would complain so I moved them.

`css-utils` can definitely be moved back to a dev dep, and we agreed that bundling the sprinkles runtime makes sense here. https://github.com/seek-oss/playroom/pull/508

PR #487, opened 2026-05-29
Inspect: Add inspector to connect rendered element to code line

michaeltaranto:
Introducing an "Inspect element" feature, that allows users to interrogate the rendered content and associate it back to the line of code in Playroom that generated it. Selecting an element will jump to the point in the code (opening the editor if not visible).

Mouse over an element to inspect demo

https://github.com/user-attachments/assets/9b6407f5-2fe8-4758-874c-2bc6bb04242a

### DIFF
--- a/cypress/support/utils.ts
+++ b/cypress/support/utils.ts
@@ -566,7 +566,6 @@ export const assertFrameSettingsCount = (count: number) =>
   cy
     .findAllByRole('button', {
       name: frameSettingsButtonLabel,
-      hidden: true,
     })
     .should('have.length', count);
 
@@ -574,7 +573,6 @@ export const toggleFrameSettingsForFrameIndex = (index: number) =>
   cy
     .findAllByRole('button', {
       name: frameSettingsButtonLabel,
-      hidden: true,
     })
     .eq(index)
     .click();
@@ -590,3 +588,67 @@ export const assertFrameSetting = (settingName: string, value: string) => {
 export const selectFrameSettingForFrameIndex = (settingName: string) => {
   cy.findByRole('menuitemcheckbox', { name: settingName }).click();
 };
+
+export const getInspectButton = () =>
+  cy.findAllByRole('button', { name: 'Inspect element' }).first();
+
+export const toggleInspectMode = (options: {
+  source: 'button' | 'keyboard';
+}) => {
+  switch (options.source) {
+    case 'button': {
+      getInspectButton().click();
+      break;
+    }
+    case 'keyboard': {
+      cy.get('body').type(cmdPlus('shift+c'));
+      break;
+    }
+    default: {
+      throw new Error('No source provided');
+    }
+  }
+};
+
+export const assertInspectModeActive = () =>
+  getInspectButton().should('have.attr', 'aria-pressed', 'true');
+
+export const assertInspectModeInactive = () =>
+  getInspectButton().should('have.attr', 'aria-pressed', 'false');
+
+export const simulateInspectMessage = (
+  type: 'hover' | 'select' | 'exit',
+  line?: number | null
+) =>
+  cy.window().then((win) => {
+    win.postMessage(
+      { source: 'Playroom Inspect', type, ...(type !== 'exit' && { line }) },
+      '*'
+    );
+  });
+
+export const assertInspectOverlayInFrame = () =>
+  getFrames()
+    .first()
+    .its('0.contentDocument.body')
+    .find('[data-testid="inspect-overlay"]')
+    .should('exist');
+
+export const getFirstFrameBody = () =>
+  getFrames().first().its('0.contentDocument.body');
+
+export const hoverTargetInFrame = (selector: string) =>
+  getFirstFrameBody()
+    .find(selector)
+    .then(($target) => {
+      const rect = $target[0].getBoundingClientRect();
+      const clientX = rect.left + rect.width / 2;
+      const clientY = rect.top + rect.height / 2;
+
+      cy.wrap($target[0].ownerDocument.body)
+        .find('[data-testid="inspect-overlay"]')
+        .trigger('mousemove', { clientX, clientY });
+    });
+
+export const getInspectHighlight = () =>
+  getFirstFrameBody().find('[data-testid="inspect-overlay"]').prev(); // Get the highlight element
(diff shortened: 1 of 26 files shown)


src/css/palettes.ts
```diff
@@ -225,6 +225,7 @@ export const light = {
     tooltip: lightPalette.gray[1000],
     textSelection: lightPalette.blue[400],
     previewBanner: surfaceLight,
+    inspectOverlay: 'hsla(209, 100%, 80%, .5)', // Blue500 at 50% opacity
```
askoufis: Any reason not to use `transparentize` here instead:
```suggestion
    inspectOverlay: transparentize(0.5, lightPalette.blue[500]), // Blue500 at 50% opacity
```

src/css/palettes.ts
```diff
@@ -275,6 +276,7 @@ export const dark = {
     tooltip: 'hsla(0,0%,0%,1)',
     textSelection: darkPalette.blue[400],
     previewBanner: surfaceLight,
+    inspectOverlay: 'hsla(209, 100%, 80%, .5)', // Blue500 at 50% opacity
```
askoufis: These are the dark tokens. but this value is from the light palette's blue500 value. Was that intentional?

src/components/Header/Header.tsx
```diff
@@ -317,6 +319,20 @@ const HeaderMenu = ({ onShareClick }: { onShareClick: () => void }) => {
...
         </Menu>
 
+        <MenuItem
+          icon={Crosshair}
+          onClick={() =>
+            dispatch({
+              type: inspectMode ? 'disableInspectMode' : 'enableInspectMode',
+            })
+          }
+          disabled={!hasCode}
```
askoufis: It feels a bit weird to disable this menu item but not disable the inspect button on the frames. Is it even worth disabling this in the menu when the inspect button still works (even though you have nothing to inspect)?

src/components/Header/Header.tsx
```diff
@@ -23,6 +23,7 @@ import {
   Eye,
   EyeClosed,
   ChevronDown,
+  Crosshair,
```
askoufis: Any specific reason why this icon was chosen? It matches safari, but I'd argue https://lucide.dev/icons/square-dashed-mouse-pointer or https://lucide.dev/icons/square-mouse-pointer would be a better fit as they're what chrome and firefox use respectively, so would likely be more recognizable.

cypress/support/utils.ts
```diff
@@ -590,3 +588,67 @@ export const assertFrameSetting = (settingName: string, value: string) => {
...
+    win.postMessage(
+      { source: 'Playroom Inspect', type, ...(type !== 'exit' && { line }) },
+      '*'
+    );
+  });
+
+export const assertInspectOverlayInFrame = () =>
+  getFrames()
+    .first()
+    .its('0.contentDocument.body')
```
askoufis:
```suggestion
  getFirstFrameBody()
```

src/components/Frame/frameMessenger.tsx
```diff
@@ -116,3 +116,61 @@ export const ScreenshotMessageReceiver = () => {
...
+const PlayroomInspectSource = 'Playroom Inspect';
+
+export const inspectMessageSender = ({
+  messageWindow,
+  action,
+}: {
+  messageWindow: Window;
+  action: 'enable' | 'disable';
+}) => {
+  messageWindow.postMessage({ source: PlayroomInspectSource, action }, '*');
```
askoufis: Does the target origin have to be `*`? I would've expected the message to be sent and received on the same origin, so wouldn't the default value (only the same origin) work?

src/components/Frame/InspectOverlay.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+    setHighlightRect(getVisibleRect(target.element));
+
+    if (target.line !== lastLineRef.current) {
+      lastLineRef.current = target.line;
+      window.parent.postMessage(
+        { source: PlayroomInspectSource, type: 'hover', line: target.line },
+        '*'
+      );
+    }
```
askoufis, line 190: Feel like maybe there's an abstraction here that combines `postMessage` with the `lastLineRef` handling. But also a bit of duplication for now is acceptable.

src/configModules/useScope.ts
```diff
@@ -8,6 +8,19 @@ import {
...
 import components from './components';
 
+const inspectCreateElement = (type: any, props: any, ...children: any[]) => {
+  if (props && props.__source) {
+    const { __source, __self, ...restProps } = props;
+    const line = __source.lineNumber - 1;
+    const lineProp =
+      typeof type === 'string'
+        ? { 'data-playroomline': line }
+        : { __playroomLine: line };
```
askoufis: Wouldn't a data attribute work for both elements and components? Why is a separate prop necessary?

src/components/Frame/InspectOverlay.tsx
```diff
@@ -0,0 +1,282 @@
...
+}
+
+function getLineFromFiber(fiber: any): number | null {
+  const props = fiber.memoizedProps;
+  const line =
+    typeof fiber.type === 'string'
+      ? props?.['data-playroomline']
+      : props?.__playroomLine;
+  return typeof line === 'number' ? line : null;
+}
```
askoufis, line 41: I'm slightly concerned about coupling to react internals. Does this work with React 18? Should we be running lockfile maintenance (or just react updates for now) so we can catch bugs sooner rather than later?

src/css/palettes.ts
michaeltaranto: It was intentional to have the same colour originally, but where the colour landed i can use the correct palette for the colour mode now 👍

src/components/Header/Header.tsx
michaeltaranto: Love it! Good find.

src/components/Frame/frameMessenger.tsx
michaeltaranto: Yeah correct. Hangover from the POC.

src/components/Frame/InspectOverlay.tsx
michaeltaranto: I forgot to role these back into the Inspect Messenger once it got to fully functional, I'll do that.

askoufis (commented):
Newer commits look good. Only suggestion would be to add some cypress tests for the `dangerouslySetInnerHTML` behaviour.

askoufis (approved):

PR #494, opened 2026-07-15
deps: Fix `typescript` dep range

askoufis:
[TypeScript v7 has no programmatic API yet](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/#:~:text=it%20does%20not%20ship%20with%20an%20API), but we rely on it in `getStaticTypes.js`. The API is slated for 7.1, but will definitely have a different shape to the existing API. Until then, we should restrict our compatible version range to reflect this, as any consumer that resolves to v7 will run into build errors.

### DIFF
--- a/.changeset/flat-hornets-drum.md
+++ b/.changeset/flat-hornets-drum.md
@@ -1,0 +1,7 @@
+---
+'playroom': patch
+---
+
+Restrict `typescript` dependency range to `^5.0.0 || ^6.0.0`
+
+Playroom's `typescript` dependency range was previously `>=5.0.0`, but TypeScript's v7 release does not have a compatible API with v5 and v6. The dependency range has been adjust to accurately reflect the versions of TypeScript that Playroom is currently compatible with.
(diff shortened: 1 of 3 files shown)


.changeset/flat-hornets-drum.md
```diff
@@ -0,0 +1,7 @@
+---
+'playroom': patch
+---
+
+Restrict `typescript` dependency range to `^5.0.0 || ^6.0.0`
+
+Playroom's `typescript` dependency range was previously `>=5.0.0`, but TypeScript's v7 release does not have a compatible API with v5 and v6. The dependency range has been adjust to accurately reflect the versions of TypeScript that Playroom is currently compatible with.
```
michaeltaranto:
```suggestion
TypeScript v7 does not have a compatible API with v5 and v6.
The dependency now accurately reflects the supported versions of TypeScript.
```

michaeltaranto (approved):

PR #495, opened 2026-07-17
Fix regex for Playroom CSS issuer path

tiplutom:
The path only works on unix systems. Because windows uses `\` as path separator, the issuer regex fails to filter correctly


.changeset/olive-years-happen.md
```diff
@@ -0,0 +1,5 @@
+---
+'playroom': patch
+---
+
+Fix regex for Playroom CSS issuer path
```
michaeltaranto:
```suggestion
webpack: Ensure Vanilla Extract stylesheets are handled on Windows
```

michaeltaranto (approved):

PR #503, opened 2026-09-16
Convert `bin` and `lib` to TypeScript and ESM

askoufis:
When combined with a custom condition there will no longer be a need for the watched `tsdown` build introduced in #476.


package.json
```diff
@@ -133,14 +131,17 @@
     "@changesets/cli": "^2.28.1",
     "@octokit/rest": "^19.0.5",
     "@testing-library/cypress": "^10.1.0",
-    "@types/jest": "^29.2.4",
+    "@types/command-line-args": "^5.2.3",
+    "@types/command-line-usage": "^5.0.4",
+    "@types/jest": "^30.0.0",
     "@types/node": "^18.11.9",
+    "babel-jest": "^30.5.1",
```
askoufis, line 138: Planning to migrate to Vitest in a followup PR. This was just introduced to make it possible to test TypeScript files.

EDIT: https://github.com/seek-oss/playroom/pull/504

package.json
```diff
@@ -29,39 +29,37 @@
...
     "serve:all": "concurrently 'npm:serve:*(!all)'",
     "build-and-serve:all": "pnpm build:all && pnpm serve:all",
     "lint": "concurrently 'npm:lint:*'",
     "lint:eslint": "NODE_OPTIONS=--max_old_space_size=8192 eslint --cache .",
-    "lint:prettier": "prettier --list-different '**/*.{js,md,ts,tsx}'",
+    "lint:prettier": "prettier --list-different .",
     "lint:tsc": "tsc --noEmit",
     "lint:cypress": "tsc --project cypress/tsconfig.json",
-    "format": "pnpm lint:eslint --fix && prettier --write '**/*.{js,md,ts,tsx}'",
+    "format": "pnpm lint:eslint --fix && prettier --write .",
```
askoufis, line 50: Explicit file extensions aren't necessary: ignores handle anything we don't want to format. The only formatting change caused by this was in `template.html`.

lib/start.js
```diff
@@ -1,25 +1,21 @@
...
     { ...config, baseUrl: '' },
-    {
-      production: false,
-      infrastructureLogging: {
-        level: 'none',
-      },
-      stats: {
-        errorDetails: true,
-      },
-    },
```
askoufis, line 18: This config was added in https://github.com/seek-oss/playroom/commit/ce918e132f063bc72d7dca625f454a4b086bcb61, but only `production` is forwarded into `makeWebpackConfig`, so they never had any effect. IMO the defaults are better than these values anyway.

bin/cli.mts
askoufis, line 1: Files are all `.mts` for now so they can be treated as ESM without declaring `"type": "module"` because that breaks other things.

michaeltaranto (approved):

